PR #27, opened 2015-02-15
Mention CIDER support

bbatsov:
CIDER 0.9 has `cljfmt` support (although it uses the library directly instead of leveraging the lein plugin).


weavejester:
Excellent news.

PR #35, opened 2015-03-13
Anchor file-pattern regex so as to not match .swp

rbxbx:
cljfmt would on occasion not play nicely with editor tempfiles. I see no reason not to anchor the regex, but perhaps I'm missing a usecase?


weavejester:
I think this was just a typo. :)

rbxbx:
such is the danger with regex!

Thanks! :)

PR #43, opened 2015-07-01
Don't add whitespace on blank lines between comments

pcarlisle:


weavejester:
The tests appear to be failing for this patch.

pcarlisle:
Ugh. Looks like I accidentally saved an unintended change to the test string right before I pushed this. Should be fixed.

weavejester:
Looks good.

PR #53, opened 2016-02-12
Add clojurescript support

### LINKED ISSUE #52: Clojurescript support
Hi great tool.

With the very fast startup time of node, a clojurescript (i.e. cljc/reader conditional) port would allow some interesting scripting possibilities with cljfmt. Personally, I'd love to use cljfmt for [my neovim plugin](https://github.com/snoe/clj-refactor.nvim).

I wouldn't mind working on a PR to attempt to add cljs support if you're interested in supporting it. 

Cheers.

snoe:
Closes #52 

Some notes:
- tests in both clj and cljs can be run with `lein test`
- I'm not 100% sure what `z/value` was doing - it's deprecated in both clj and cljs but I think with the `when` guard just using `sexpr` is safe.
- The macro inlines the resources directly into javascript.
- Since the cljs rewrite port has functions in slightly different namespaces I used a combination of `def` and `refer` to minimize the number of reader conditionals.

### DIFF
--- a/cljfmt/project.clj
+++ b/cljfmt/project.clj
@@ -4,5 +4,18 @@
   :scm {:dir ".."}
   :license {:name "Eclipse Public License"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
-  :dependencies [[org.clojure/clojure "1.6.0"]
-                 [rewrite-clj "0.4.12"]])
+  :dependencies [[org.clojure/clojure "1.7.0"]
+                 [org.clojure/clojurescript "1.7.228"]
+                 [rewrite-clj "0.4.12"]
+                 [rewrite-cljs "0.4.0"]]
+  :plugins [[lein-cljsbuild "1.1.2"]]
+  :hooks [leiningen.cljsbuild]
+  :cljsbuild {:builds
+              {"dev" {:source-paths ["src" "test"]
+                      :compiler {:main cljfmt.test-runner
+                                 :output-to "target/out/tests.js"
+                                 :output-dir "target/out"
+                                 :target :nodejs
+                                 :optimizations :none}}}
+              :test-commands
+              {"dev" ["node" "target/out/tests.js"]}})
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -1,11 +1,33 @@
 (ns cljfmt.core
-  (:import [clojure.lang Symbol]
-           [java.util.regex Pattern])
-  (:require [clojure.java.io :as io]
-            [clojure.zip :as zip]
-            [rewrite-clj.parser :as p]
-            [rewrite-clj.node :as n]
-            [rewrite-clj.zip :as z]))
+  #?@(:clj
+       [(:require
+         [clojure.java.io :as io]
+         [clojure.zip :as zip]
+         [rewrite-clj.node :as n]
+         [rewrite-clj.parser :as p]
+         [rewrite-clj.zip :as z
+          :refer [append-space edn skip whitespace-or-comment?]])
+        (:import java.util.regex.Pattern)]
+       :cljs
+       [(:require
+         [cljs.reader :as reader]
+         [clojure.zip :as zip]
+         [rewrite-clj.node :as n]
+         [rewrite-clj.parser :as p]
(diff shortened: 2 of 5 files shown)


weavejester:
Thanks! I'll review.

cljfmt/project.clj
```diff
@@ -1,8 +1,21 @@
-(defproject cljfmt "0.3.0"
+(defproject cljfmt "0.4.0"
```
weavejester: Pull requests shouldn't change the version number of the project.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -136,16 +158,23 @@
...
 
+(defn pattern?
+  [v]
+  (instance? #?(:clj Pattern :cljs js/RegExp) v))
+
 (defn- indent-matches? [key sym]
-  (condp instance? key
-    Symbol  (= key sym)
-    Pattern (re-find key (str sym))))
+  (cond
```
weavejester: Could maybe make this a little more concise, e.g.

```clojure
(cond
  (symbol? key)  (= key sym)
  (pattern? key) (re-find key (str sym)))
```

cljfmt/src/cljfmt/macros.clj
```diff
@@ -0,0 +1,7 @@
+(ns cljfmt.macros
+  (:require [clojure.java.io :as io]))
+
+(defmacro read-resource
```
weavejester: We could do without this file if we just used a different name for the macro. So in the core namespace:

```clojure
#?(:clj (def read-resource (comp read-string slurp io/resource)))
#?(:clj (defmacro read-resource* [path] `'(~read-resource path)))
```

I also think you can avoid the `read-string` and `pr-str` if you quote inside the syntax quote, but I could be wrong about this.

Also, are there any disadvantages in using the macro in *both* Clojure and ClojureScript? Because Clojure compiles top-level forms, it shouldn't matter whether we use `read-resource` or `read-resource*` in a `def`, as far as I can tell...

snoe:
Pushed the changes, I opted to use the macro for both as I believe you're right that there's really no disadvantage.

weavejester:
Great, thanks!

Could you add ClojureScript testing to the `.travis.yml` file? I think you'll need to add a new `script` key.

snoe:
So it looks like travis is already running the clojurescript tests (`lein test` will run both). Is that not sufficient?

https://travis-ci.org/weavejester/cljfmt/builds/108830160
```
lein test user
Ran 0 tests containing 0 assertions.
0 failures, 0 errors.
Running ClojureScript test: dev
Testing cljfmt.core-test
Ran 6 tests containing 74 assertions.
0 failures, 0 errors.
The command "lein test" exited with 0.```

**Comment by weavejester** [comment]
It doesn't look like it's running both sets of tests. Perhaps because it's using Leiningen 1.5.1? I'll investigate.

**Comment by weavejester** [comment]
It seems that Travis is *still* using a version of Leiningen that doesn't support `cljc` files, which should have really been fixed last summer. It looks like Travis are not exactly quick at updating their build environments!

It should be possible to get around this by changing the `before_install` option to `lein upgrade && cd cljfmt`. I'm going to test it in a separate branch, but Travis is also being slow to run the build (over 30 minutes so far). At this point I'm tempted to try out CircleCI instead.

**Comment by weavejester** [comment]
Here's the `.travis.yml` file that eventually worked for me. Can you add it to your PR so the CI checks it correctly?

```yaml
language: clojure
before_install: yes | sudo lein upgrade && cd cljfmt
```

snoe:
Much better, thanks.

```
lein test cljfmt.core-test
Ran 6 tests containing 74 assertions.
0 failures, 0 errors.
Running ClojureScript test: dev
Testing cljfmt.core-test
Ran 6 tests containing 74 assertions.
0 failures, 0 errors.
The command "lein test" exited with 0.
```

weavejester:
Great! Could you squash your commits? Then I think I'm good to merge.

snoe:
There you go, all squashed.

snoe:
@weavejester any chance this could be released on clojars?

weavejester:
Released as 0.4.0. Sorry for the delay.

PR #54, opened 2016-02-24
Standalone check and format

### LINKED ISSUE #50: Could not find project.clj
I'm trying to format a single file by running `lein cljfmt check <filename>` but it keeps telling me that it `Couldn't find project.clj, which is needed for cljfmt`. Why does it need a `project.clj` even for a single file?

arrdem:
This changeset tweaks cljfmt so that it can be run outside of a Leiningen project so long as one or more source paths are provided.

Fixes #50

### DIFF
--- a/lein-cljfmt/src/leiningen/cljfmt.clj
+++ b/lein-cljfmt/src/leiningen/cljfmt.clj
@@ -36,7 +36,7 @@
       (.getPath)))
 
 (defn project-path [project file]
-  (relative-path (io/file (:root project)) (io/file file)))
+  (relative-path (io/file (:root project ".")) (io/file file)))
 
 (defn format-diff [project file]
   (let [filename (project-path project file)
@@ -57,7 +57,9 @@
 
 (defn check
   ([project]
-   (apply check project (format-paths project)))
+   (if project
+     (apply check project (format-paths project))
+     (main/abort "No project found and no source paths provided")))
   ([project path & paths]
    (let [files   (mapcat (partial find-files project) (cons path paths))
          invalid (remove (partial valid-format? project) files)]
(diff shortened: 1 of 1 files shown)


lein-cljfmt/src/leiningen/cljfmt.clj
```diff
@@ -57,8 +57,12 @@
 
 (defn check
   ([project]
-   (apply check project (format-paths project)))
+   (if project
+     (apply check project (format-paths project))
+     (main/abort "No project found and no source paths provided")))
   ([project path & paths]
+   (when-not project
+     (main/info "Warning: project is nil, assuming . is project root"))
```
weavejester: Instead of `main/info`, perhaps `main/warn` would be more appropriate.

However, do we need this warning? I don't think we need to print a warning message if the user isn't using the plugin in an unexpected way. Is there a reason we'd want to keep the warning?

weavejester:
Thanks for the patch! I only have one comment on the code itself.

arrdem:
Sure! I'm motivated by getting the CLJ part of the Clojarr source under the watchful eyes of the evil whitespace bots :stuck_out_tongue: one sec and I'll squash a fix in.

weavejester:
Did you read the second paragraph of my comment?

arrdem:
Sorry, I didn't see that there was more. 

I agree that it's a common enough default that it probably shouldn't surprise anyone, but probably is the keyword. I don't think having this warning hurts in any meaningful way, it's only printed once per task rather than once per file or something much noisier and there's a chance it'll help someone. I think it's worth having just because of its low price and that off chance.

weavejester:
Under what situation would it be useful, though? I can see it getting annoying if someone happens to use it often on a project without a `project.clj` file, so I think there needs to be a reason to put it in.

arrdem:
After some pondering you're right there isn't a situation when that warning matters, because if the user provides a relative path to `.` it'll work "as expected", an absolute path will also work "as expected", the only failure mode is if the user provides a path relative to something other than `.`, which doesn't make sense for us to support anyway.

I'll remove it.

weavejester:
Thanks! Can you squash your commits and change the commit message to something like:

    Allow source files to be checked outside of Lein projects

arrdem:
Done.

PR #56, opened 2016-02-24
Format once

### LINKED ISSUE #55: Reformats files twice
`valid-format?` calls `reformat-string` and checks `=` to see if anything changed, but when computing patches or reformatting files `reformat-string` is called again either to generate a diff or just for output. This means that large files get slurped and reformatted not once but twice for no particular reason.

arrdem:
This patch fixes #55, by inlining the `valid-format?` check and tweaking diff generation so that files are never read or formatted twice. This gives the expected 50% speedup.

before:
> lein cljfmt fix src/clj  772.30s user 1.78s system 101% cpu 12:41.43 total

after:
> lein cljfmt fix src/clj  396.35s user 1.40s system 103% cpu 6:23.34 total

merges cleanly with #54

### DIFF
--- a/lein-cljfmt/src/leiningen/cljfmt.clj
+++ b/lein-cljfmt/src/leiningen/cljfmt.clj
@@ -38,14 +38,16 @@
 (defn project-path [project file]
   (relative-path (io/file (:root project)) (io/file file)))
 
-(defn format-diff [project file]
-  (let [filename (project-path project file)
-        original (slurp (io/file file))
-        revised  (reformat-string project original)
-        diff     (diff/unified-diff filename original revised)]
-    (if (get-in project [:cljfmt :ansi?] true)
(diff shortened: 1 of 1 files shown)


weavejester:
Thanks for the patch! Looking at your commits, it seems like 79307f2 has no purpose without 12052b7, so I think those two should be squashed together.

So perhaps we start with 4df8fc6, but let's change the commit message to simply:

    Don't reformat files twice in fix

The "how" is obvious from the diff, I think, so there's no need to explain it in the commit message.

Then we combine 79307f2 and 12052b7 into one, perhaps with the message:

    Don't reformat files twice in check

Again, I think the diff explains the "how".

You also should remove the `valid-format?` function, since it's no longer necessary after 12052b7.

arrdem:
Done and done.

PR #62, opened 2016-03-06
Handle file parse errors more gracefully

arrdem:
This patch serves to inform the user which file was to blame if a parse
error is encountered.

### DIFF
--- a/lein-cljfmt/src/leiningen/cljfmt.clj
+++ b/lein-cljfmt/src/leiningen/cljfmt.clj
@@ -3,6 +3,7 @@
   (:require [cljfmt.core :as cljfmt]
             [clojure.java.io :as io]
             [clojure.string :as str]
+            [clojure.stacktrace :refer [print-stacktrace]]
             [leiningen.core.main :as main]
             [leiningen.cljfmt.diff :as diff]
             [meta-merge.core :refer [meta-merge]]))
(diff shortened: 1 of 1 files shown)


lein-cljfmt/src/leiningen/cljfmt.clj
```diff
@@ -59,20 +59,28 @@
...
-             :let  [original (slurp f)
-                    revised  (reformat-string project original)]
-             :when (not= original revised)]
-       (main/warn (project-path project f) "has incorrect formatting:")
-       (main/warn (format-diff project f original revised))
-       (swap! flag inc))
-     (if (zero? @flag)
+   (let [files             (mapcat (partial find-files project) (cons path paths))
+         incorrect-counter (atom 0)
+         failure-counter   (atom 0)]
```
weavejester: It would be nice to change this `doseq` into a `reduce` or `loop` sometime, but for now it's probably okay.

arrdem: I don't think that `reduce` is appropriate unless this is reworked to return some sort of structure which returns counters rather than using local mutation. `loop` would be worse than `doseq` because at least `doseq` gives you the loop control structure for free.

lein-cljfmt/src/leiningen/cljfmt.clj
```diff
@@ -59,20 +59,28 @@
...
+     (doseq [^java.io.File f files]
+       (let [original (slurp f)]
+         (try
+           (let [revised (reformat-string project original)]
+             (when (not= original revised)
+               (main/warn (project-path project f) "has incorrect formatting:")
+               (main/warn (format-diff project f original revised))
+               (swap! incorrect-counter inc)))
+           (catch Exception e
+             (main/warn "Failed to format file:" (.getPath f))
```
weavejester: Instead of `(.getPath f)`, `(project-path project f)` would produce a more concise filepath.

I also think we might want to print out the exception and stacktrace as well. So maybe:

```clojure
(main/warn (project-path project f) "could not be formatted:")
(binding [*out* *err*] (clojure.stacktrace/print-stack-trace e))
```

lein-cljfmt/src/leiningen/cljfmt.clj
weavejester: > I don't think that `reduce` is appropriate unless this is reworked to return some sort of structure which returns counters rather than using local mutation.

Right. That's what I had in mind.

arrdem:
Okay. Stack traces will now be printed, `project-path` paths will be used, and I rewrite `check` using a loop and a helper kinda in the style of the `clojure.test` status accumulator. Whether that's overkill or not I'll leave to your judgement.

weavejester:
Sorry for the delay in reviewing this. This looks fine - can you squash your commits?

arrdem:
Willco. No sweat, I've got plenty of other places to get my daily commits in :D

arrdem:
Squashed.

weavejester:

```
Don't fail on first unreadable file

If lein-cljfmt fails to read a file, print an error and continue
processing. After formatting print the total number of errors in
addition to files changed.
```

I think it's fairly obvious why this is a better approach than failing on the first error :)

arrdem:
Done.

weavejester:
Perfect, thanks!

arrdem:
:shipit:

weavejester:
It looks like I merged too soon. I assumed that you had it working, but it doesn't compile. I'll see if I can fix the errors.

arrdem:
Apologies :|

PR #69, opened 2016-03-28
Add option to remove trailing whitespace

tomjkidd:
The implementation respects the surrounding whitespace and handles whitespace at the end of the string for cases where I would expect it is desired. Note that I am new to the codebase, open to feedback if I missed something.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -273,6 +273,19 @@
   ([form indents]
    (indent (unindent form) indents)))
 
+(defn root? [zloc]
+  (nil? (zip/up zloc)))
+
+(defn final? [zloc]
+  (and (nil? (zip/right zloc)) (root? (zip/up zloc))))
+
+(defn- trailing-whitespace? [zloc]
+ (and (whitespace? zloc)
+      (or (zlinebreak? (zip/right zloc)) (final? zloc))))
+
+(defn remove-trailing-whitespace [form]
+  (transform form edit-all trailing-whitespace? zip/remove))
+
 (defn reformat-form
   [form & [{:as opts}]]
   (-> form
@@ -283,7 +296,9 @@
       (cond-> (:insert-missing-whitespace? opts true)
         insert-missing-whitespace)
       (cond-> (:indentation? opts true)
-        (reindent (:indents opts default-indents)))))
+        (reindent (:indents opts default-indents)))
+      (cond-> (:remove-trailing-whitespace? opts true)
+        remove-trailing-whitespace)))
 
 (defn reformat-string [form-string & [options]]
   (-> (p/parse-string-all form-string)
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -172,6 +172,26 @@
   (is (= (reformat-string "(foo)\n;bar\n;baz\n;qux\n(bang)")
          "(foo)\n;bar\n;baz\n;qux\n(bang)")))
 
+(deftest test-trailing-whitespace
+  (testing "trailing-whitespace"
+    (is (= (reformat-string "(foo bar) ")
+           "(foo bar)"))
+    (is (= (reformat-string "(foo bar)\n")
+           "(foo bar)\n"))
+    (is (= (reformat-string "(foo bar) \n ")
+           "(foo bar)\n"))
+    (is (= (reformat-string "(foo bar) \n(foo baz)")
+           "(foo bar)\n(foo baz)"))
+    (is (= (reformat-string "(foo bar)\t\n(foo baz)")
+           "(foo bar)\n(foo baz)")))
+
+  (testing "preserve surrounding whitespace"
+    (is (= (reformat-string "( foo bar ) \n" {:remove-surrounding-whitespace? false})
+           "( foo bar )\n"))
+    (is (= (reformat-string
+            "( foo bar )   \n( foo baz )\n" {:remove-surrounding-whitespace? false})
+           "( foo bar )\n( foo baz )\n"))))
+
 (deftest test-options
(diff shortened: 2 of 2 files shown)


weavejester:
Thanks for the patch! It's well done for someone who's new to the codebase (and to Clojure). However, I do have a few suggestions which I'll add as inline comments on the commit.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -273,6 +274,21 @@
   ([form indents]
    (indent (unindent form) indents)))
 
+(defn- trailing-whitespace? [zloc]
```
weavejester: I think this function is a little overly complex. What we want is to match two circumstances:

1. Whitespace followed by a newline
2. Whitespace that's at the end of the file

That sound about right? So we could write (1) as:

```clojure
(defn- trailing-whitespace? [zloc]
  (let [r (zip/right zloc)]
    (and (whitespace? zloc) (line-break? r) (not (comment? r))))
```

As you likely discovered for yourself, `line-break?` matches both newlines and comments:

```clojure
(defn- line-break? [zloc]
  (or (zlinebreak? zloc) (comment? zloc)))
```

That was because I didn't think there'd be cause to treat them differently. However, in this case we do want to treat them differently, so we can just use the raw `zlinebreak?` function (which probably needs a better name, now).

```clojure
(defn- trailing-whitespace? [zloc]
  (and (whitespace? zloc) (zlinebreak? (zip/right zloc))))
```

That takes care of (1). Now we just need to figure out (2). It turns out that if we just use:

```clojure
(nil? (zip/right zloc))
```

Then we fall afoul of whitespace internal to forms, e.g. `( foo )` becomes `( foo)`. I see you noticed this as well, because you added in `(not (surrounding? zloc whitespace?))`.

However, I think there's a better way of handling this. We check `(nil? (zip/right zloc))` only when we know we're directly below the root form:

```clojure
(defn final? [zloc]
  (and (nil? (zip/right zloc)) (root? (zip/up zloc))))
```

How do we know what the root form is? Well, when we can no longer go `zip/up`:

```clojure
(defn root? [zloc]
  (nil? (zip/up zloc)))
```

So all together:

```clojure
(defn- trailing-whitespace? [zloc]
  (and (whitespace? zloc)
       (or (zlinebreak? (zip/right zloc)) (final? zloc))))
```

This also removes the need for `trimr` in `reformat-string`.

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -172,6 +172,25 @@
   (is (= (reformat-string "(foo)\n;bar\n;baz\n;qux\n(bang)")
          "(foo)\n;bar\n;baz\n;qux\n(bang)")))
 
+(deftest test-trailing-whitespace
+  (testing "trailing-whitespace"
+    (is (= (reformat-string "(foo bar) ")
+           "(foo bar)"))
+    (is (= (reformat-string "(foo bar)\n")
+           "(foo bar)"))
```
weavejester: Having a newline at the end of the file is valid, so `"(foo bar)\n"` shouldn't be changed.

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -172,6 +172,25 @@
...
+    (is (= (reformat-string "(foo bar)   \n  ")
```
weavejester: `"(foo bar)   \n  "` should reformat to `"(foo bar)   \n"`.

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -172,6 +172,25 @@
...
+           "(foo bar)"))
+    (is (= (reformat-string "(foo bar) \n(foo baz)")
+           "(foo bar)\n(foo baz)"))
+    (is (= (reformat-string "(foo bar)\t\n(foo baz)")
+           "(foo bar)\n(foo baz)")))
+
+  (testing "preserve surrounding whitespace"
+    (is (= (reformat-string "( foo bar )   \n" {:remove-surrounding-whitespace? false})
+           "( foo bar )"))
+    (is (= (reformat-string "( foo bar )   \n( foo baz )\n" {:remove-surrounding-whitespace? false})
```
weavejester: This line was a little long. Perhaps break it up a bit:

```clojure
(is (= (reformat-string
        "( foo bar )   \n( foo baz )\n" {:remove-surrounding-whitespace? false})
       "( foo bar )\n( foo baz )"))))
```

weavejester:
Your commit message also needs some minor changes. I follow [the seven rules](http://chris.beams.io/posts/git-commit/#seven-rules), so your commit message should ideally read:

    Add remove-trailing-whitespace functionality

tomjkidd:
Thank you for the details in how you got from my code to a cleaner solution. I really appreciate the feedback you gave and the link for commit messages. I am looking into how to rebase the commits so that they are in a cleaner format, let me know if you see anything else that needs to be done.

weavejester:
You're welcome :)

For rebasing the commits, you'll want to perform an [interactive rebase](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html). If you could squash all the changes into one commit, then your PR will be ready to merge.

tomjkidd:
That link was helpful, too! Thanks again.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -3,6 +3,7 @@
        [(:require
          [clojure.java.io :as io]
          [clojure.zip :as zip]
+         [clojure.string :as str]
```
weavejester: This can be removed now, right?

tomjkidd: I was in the process of my second interactive rebase having noticed that, should be good now.

weavejester:
Thanks! I'll do a release shortly.

weavejester:
Released 0.5.0. Do you have a Twitter handle? I can mention you on the release tweet.

tomjkidd:
Oh, don't worry about it! Thanks for the opportunity.

PR #70, opened 2016-04-14
Use exit code to indicate check problems

maio:
Exit with code 1 if there is a file with incorrect formatting.
Exit with code 2 if there were some errors.

Useful if you want to perform some action depending on check
results (e.g. trigger build failure).

### DIFF
--- a/lein-cljfmt/src/leiningen/cljfmt.clj
+++ b/lein-cljfmt/src/leiningen/cljfmt.clj
@@ -84,6 +84,12 @@
       (main/warn path "has incorrect formatting")
       (main/warn diff))))
 
+(defn exit [counts]
+  (when-not (zero? (:error counts 0))
+    (main/exit 2))
+  (when-not (zero? (:incorrect counts 0))
+    (main/exit 1)))
+
 (defn print-final-count [counts]
   (let [error     (:error counts 0)
         incorrect (:incorrect counts 0)]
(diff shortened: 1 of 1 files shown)


lein-cljfmt/src/leiningen/cljfmt.clj
```diff
@@ -84,6 +84,12 @@
       (main/warn path "has incorrect formatting")
       (main/warn diff))))
 
+(defn exit [counts]
+  (when-not (zero? (:incorrect counts 0))
+    (leiningen.core.main/exit 1))
```
weavejester: Why not `(main/exit 1)` here?

lein-cljfmt/src/leiningen/cljfmt.clj
```diff
@@ -105,15 +111,16 @@
...
-   (->> (cons path paths)
-        (transduce
-         (comp (mapcat (partial find-files project))
-               (map (partial check-one project))
-               (map (fn [status]
-                      (print-file-status project status)
-                      (:counts status))))
-         (completing merge-counts))
-        (print-final-count))))
+   (let [counts (->> (cons path paths)
```
weavejester: You could factor out the `->>` part and just put `(cons path paths)` on the end of the transduce.

maio:
@weavejester thanks for comments. They should be fixed now.

weavejester:
Perfect, thanks :)

weavejester:
Just tested it and it all works, but I notice that if there are errors *and* incorrect formatting, it returns an error code of 1. I think errors are more important, so it should return 2 in that case. Do you agree?

maio:
:+1:

weavejester:
Then if you swap the `when-not` statements in the `exit` function, I think it will be ready to merge.

maio:
It should be all fixed now.

weavejester:
Perfect, thanks!

PR #80, opened 2016-09-16
Don't add whitespace after reader conditionals

jetmind:
This updates `rewrite-clj` & `rewrite-cljs` to newest versions and restricts missing whitespace check for reader conditionals.

### DIFF
--- a/cljfmt/project.clj
+++ b/cljfmt/project.clj
@@ -6,8 +6,8 @@
             :url "http://www.eclipse.org/legal/epl-v10.html"}
   :dependencies [[org.clojure/clojure "1.7.0"]
                  [org.clojure/tools.reader "1.0.0-alpha4"]
-                 [rewrite-clj "0.4.12"]
-                 [rewrite-cljs "0.4.1"]]
+                 [rewrite-clj "0.5.2"]
+                 [rewrite-cljs "0.4.2"]]
   :plugins [[lein-cljsbuild "1.1.2"]]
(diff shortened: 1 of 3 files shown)


weavejester:
Thanks! Could you also add a test? Something that would fail before, but is fixed with the changes you've made. I'm also a little unclear on the reason why the missing whitespace check doesn't apply to conditionals.

jetmind:
Added a test. The test fails on master as following:

```
FAIL in (missing-whitespace-reader-conditionals) (core_test.cljc:162)
expected: (= (reformat-string "#?(:cljs(bar 1) :clj(foo 2))") "#?(:cljs (bar 1) :clj (foo 2))")
  actual: (not (= "#? (:cljs (bar 1) :clj (foo 2))" "#?(:cljs (bar 1) :clj (foo 2))"))

FAIL in (missing-whitespace-reader-conditionals) (core_test.cljc:164)
expected: (= (reformat-string "#?@(:cljs[foo bar] :clj[baz quux])") "#?@(:cljs [foo bar] :clj [baz quux])")
  actual: (not (= "#? @(:cljs [foo bar] :clj [baz quux])" "#?@(:cljs [foo bar] :clj [baz quux])"))
```

You can see that it adds unneeded whitespace between `#?` and following form or between `#?` and `@` for splicing version of conditional. I hope this clarifies why this check is needed.

weavejester:
Yes, it does clarify it perfectly, thanks!

PR #85, opened 2016-10-19
Follow style guide for as->

MatthewDarling:
as-> has two special arguments: the expression being bound, and the name being introduced inside the body of the threading. This gives it the correct indentation.

Before:

```clojure
(as->> 1 $
       (inc $)
       (* 2 $))
```

After:

```clojure
(as->> 1 $
  (inc $)
  (* 2 $))
```


weavejester:

    Add block indent for as->

weavejester:
The commit still has a bunch of text in the body that I don't think is necessary. Can you remove all but the first line of the commit?

MatthewDarling:
Should be good now :)

PR #86, opened 2016-10-20
Only filter the portion of filename within project

MatthewDarling:
The current version applies the regex against the entire filename,
meaning that it can have different behaviour for people depending on
their local directory structure. By removing the parent directory name,
we make the behaviour more repeatable.

Consider a repo with this structure:

    foo/src/org/foo/bar.clj
    foo/test/org/foo/bar_test.clj

If you want to apply the formatting only to files under the src folder,
you might write this regex:

    #"src/.+\.clj[csx]?$"

We want it to indent bar.clj but not bar_test.clj.

However, imagine two different developers have this repo, and they store
their code in different directories:

    Dev 1: ~/code/foo
    Dev 2: ~/src/foo

For Dev 1, the regex above will work correctly. However, for Dev 2, the
regex will seem "broken" - bar_test.clj will still be indented! And
that's because it's matching the parent directory of src.

Removing the parent directory before matching avoids the influence of
the local directory structure on a given machine.

### DIFF
--- a/lein-cljfmt/src/leiningen/cljfmt.clj
+++ b/lein-cljfmt/src/leiningen/cljfmt.clj
@@ -8,8 +8,13 @@
             [leiningen.cljfmt.diff :as diff]
             [meta-merge.core :refer [meta-merge]]))
 
+(defn relative-path [dir file]
+  (-> (.toURI dir)
+      (.relativize (.toURI file))
+      (.getPath)))
+
 (defn grep [re dir]
-  (filter #(re-find re (str %)) (file-seq (io/file dir))))
+  (filter #(re-find re (relative-path dir %)) (file-seq (io/file dir))))
 
 (defn file-pattern [project]
   (get-in project [:cljfmt :file-pattern] #"\.clj[sx]?$"))
@@ -26,11 +31,6 @@
 
 (defn reformat-string [project s]
   (cljfmt/reformat-string s (meta-merge default-config (:cljfmt project {}))))
-
-(defn relative-path [dir file]
-  (-> (.toURI dir)
-      (.relativize (.toURI file))
-      (.getPath)))
 
 (defn project-path [project file]
   (relative-path (io/file (:root project ".")) (io/file file)))


MatthewDarling:
This is potentially a breaking change for any projects that were relying on the old behaviour. Not sure how best to communicate that, or whether that's even an acceptable change to make.

weavejester:
Thanks for the patch. Lein-cljfmt should ignore the outer directories, so this is a bug fix :)

However, have you tested this change? `remove-dir-from-filename` takes two arguments, but when you call it you only pass one.

Also, all of the description of the PR is in your commit message (I assume that's the default of some tool?). Can you remove all of the unnecessary information? The commit message just needs to describe the change, rather than provide a full justification, so you can just delete everything except the first line.

MatthewDarling:
Oops. I had tested it before creating ```remove-dir-from-filename```, but not after factoring that out into a separate function.

Should be fixed now, both the code and the commit message.

I was worried this would perform badly, so I benchmarked before & after with [Criterium](https://github.com/hugoduncan/criterium). In the results below, ```grep``` is the current version and ```smarter-grep``` is my changed version:

```clojure
leiningen.cljfmt> (require 'criterium.core)
;; => nil

;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;; new test ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
leiningen.cljfmt> (criterium.core/bench (grep #"src/.+\.clj[sx]?$" "/Users/matthewdarling/src/myproject"))
WARNING: Final GC required 5.584298539364606 % of runtime
Evaluation count : 886981140 in 60 samples of 14783019 calls.
             Execution time mean : 64.296780 ns
    Execution time std-deviation : 3.305557 ns
   Execution time lower quantile : 56.003595 ns ( 2.5%)
   Execution time upper quantile : 70.025426 ns (97.5%)
                   Overhead used : 1.745570 ns

Found 3 outliers in 60 samples (5.0000 %)
	low-severe	 3 (5.0000 %)
 Variance from outliers : 36.8855 % Variance is moderately inflated by outliers
;; => nil

;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;; new test ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
leiningen.cljfmt> (criterium.core/bench (smarter-grep #"src/.+\.clj[sx]?$" "/Users/matthewdarling/src/myproject"))
Evaluation count : 868663860 in 60 samples of 14477731 calls.
             Execution time mean : 60.536168 ns
    Execution time std-deviation : 3.743363 ns
   Execution time lower quantile : 55.430309 ns ( 2.5%)
   Execution time upper quantile : 68.367699 ns (97.5%)
                   Overhead used : 1.745570 ns

Found 1 outliers in 60 samples (1.6667 %)
	low-severe	 1 (1.6667 %)
 Variance from outliers : 46.7423 % Variance is moderately inflated by outliers
;; => nil

;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;; new test ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
leiningen.cljfmt> (criterium.core/bench (grep #"src/.+\.clj[sx]?$" "/Users/matthewdarling/src/myproject"))
Evaluation count : 862410960 in 60 samples of 14373516 calls.
             Execution time mean : 65.954064 ns
    Execution time std-deviation : 4.803710 ns
   Execution time lower quantile : 58.238083 ns ( 2.5%)
   Execution time upper quantile : 71.947394 ns (97.5%)
                   Overhead used : 1.745570 ns

Found 1 outliers in 60 samples (1.6667 %)
	low-severe	 1 (1.6667 %)
 Variance from outliers : 55.0903 % Variance is severely inflated by outliers
;; => nil

;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;; new test ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
leiningen.cljfmt> (criterium.core/bench (smarter-grep #"src/.+\.clj[sx]?$" "/Users/matthewdarling/src/myproject"))
Evaluation count : 907382940 in 60 samples of 15123049 calls.
             Execution time mean : 61.251390 ns
    Execution time std-deviation : 2.289370 ns
   Execution time lower quantile : 58.378686 ns ( 2.5%)
   Execution time upper quantile : 66.250485 ns (97.5%)
                   Overhead used : 1.745570 ns

Found 2 outliers in 60 samples (3.3333 %)
	low-severe	 2 (3.3333 %)
 Variance from outliers : 23.8349 % Variance is moderately inflated by outliers
;; => nil
```

Strangely, the mean is better in both runs with the new version. At least, this reassures me that the string replace didn't make it 10x slower or anything like that.

lein-cljfmt/src/leiningen/cljfmt.clj
```diff
@@ -8,8 +8,12 @@
             [leiningen.cljfmt.diff :as diff]
             [meta-merge.core :refer [meta-merge]]))
 
+(defn remove-dir-from-filename [dir filename]
+  (str/replace filename dir ""))
```
weavejester: This doesn't ensure the `dir` is removed from the beginning of the string only, right?

Perhaps something more like:

```clojure
(if (.startsWith ^String filename dir)
  (subs filename (count dir))
  filename)
```

MatthewDarling: Good call. Especially bad if someone has their project in a root directory, like:

```bash
c/myproject/src/org/...
```

well, now every "c" in the filenames will be removed.

Your version fixes that, I'll make the switch.

MatthewDarling:
Hmm, looks like I messed up the "change requested" thing with my amend + force push. I'd been doing that because it seemed like you preferred to have a single commit with a particular message.

If there's multiple commits needed during a PR, what's the right way to handle that?

weavejester:
I don't think you've messed up anything. I can still approve the changes.

My approach to PRs is to do multiple small commits while the PR is a work in progress, and then once the code is correct I'll go back and clean up the commit history, putting it either in one or several commits.

But your approach of overwriting the commit is fine too. It's not like this is a big commit, and it might save time.

weavejester:
I get an error when I test this:

```
java.lang.ClassCastException: java.io.File cannot be cast to java.lang.String
 at leiningen.cljfmt$remove_dir_from_filename.invokeStatic (cljfmt.clj:12)
    leiningen.cljfmt$remove_dir_from_filename.invoke (cljfmt.clj:11)
    leiningen.cljfmt$grep$fn__3858.invoke (cljfmt.clj:17)
    clojure.core$filter$fn__4812.invoke (core.clj:2712)
```

MatthewDarling:
How are you testing it? The test directory seems empty for ```lein-cljfmt``` so I wasn't too sure what to do.

It worked for me just now doing this:

  1. ```lein do clean, repl``` in lein-cljfmt directory
  2. Connect with CIDER from Emacs
  3. Compile ```leiningen.cljfmt``` namespace
  4. ```(grep #"src/.+\.clj[sx]?$" "/Users/matthewdarling/src/myproject/")```

There might be differences when statically compiled vs doing it in a REPL, maybe...?

One important thing is that ```grep``` converts the java.io.File to a string before calling ```remove-dir-from-filename```.

weavejester:
Just go into the directory and run:

    lein cljfmt check

MatthewDarling:
Ahh, when it's called from command line, ```dir``` is a java.io.File. It needs to be converted too.

I looked at using ```project-path``` as a replacement for ```remove-dir-from-filename```. However, it seems like the ```dir``` passed to ```grep``` might not be the project root, so it wouldn't be equivalent.

MatthewDarling:
Ahh, when it's called from command line, ```dir``` is a java.io.File. It needs to be converted too. I was calling it wrong this whole time.

The existing ```relative-path``` function does what I want, so I'll move it up in the file and reuse it.

weavejester:
Looks good. Merged.

Issue #82, opened 2016-09-30
Indentation lost after comment line


weissjeffm:
When a line is commented out and what's next is `]`, then that gets aligned at column 1 for some reason. It should be at column 4, i believe (that's what clojure-mode does).

Input for cljfmt fix:

``` clj
(let [x y]
  [this
   that
   ;;other
   ]
  )
```

output

``` clj
(let [x y]
  [this
   that
   ;;other
])
```

expected

``` clj
(let [x y]
  [this
   that
   ;;other
   ])
```

weissjeffm:
BTW the same thing happens inside a map literal also.

arrdem:
So I think what's happening here is that the comment is messing up scanning backwards to the previous form to get indentation... definitely a bug in cljfmt not rewrite-clj and shouldn't be too hard to fix.

MatthewDarling:
Stepping through with the CIDER debugger, [`remove-surrounding-whitespace`](https://github.com/weavejester/cljfmt/blob/master/cljfmt/src/cljfmt/core.cljc#L58) is responsible. The whitespace disappears after a few iterations through `edit-all`.

When it looks at the whitespace starting the line after `;;;test1`, `surrounding-whitespace?` returns true, because the current node is whitespace and [part of this `or` is returning true](https://github.com/weavejester/cljfmt/blob/master/cljfmt/src/cljfmt/core.cljc#L48).

I guess we'd want to modify that conditional on line 48 to fix the issue? I'm not sure I understand enough to do that, so far.

MatthewDarling:
Starting to understand `rewrite-clj` and `cljfmt` a bit better, so I took another look at this today.

## Slightly more informed description of the issue 

In terms of why `remove-surrounding-whitespace` is responsible: the README example is turning `( foo )` into `(foo)`. So `surrounding-whitespace?` check if a node is whitespace, and that it either has nothing to its left, or nothing but whitespace to its right.

Code like this triggers the "nothing but whitespace to the right" case (`<>` denotes the current node):

```clojure
(let [x y]
  [this
   that
   ;;other
<   >]
  )
```

There are no nodes to the right of that, so the whitespace node is removed under the assumption it's a `(foo )` situation.

## A possible fix

To distinguish between the two, we could check if the previous node is a comment, and only remove if it isn't:

```clojure
(defn- surrounding? [zloc p?]
  (and (p? zloc)
       (or (nil? (zip/left zloc))
[...]

PR #88, opened 2016-12-15
Add failing test for post-comment indentation

jcf:
Hey @weavejester,

I've noticed the logic for indenting comments nested inside a `let` within a `defn` is slightly off (as has been reported via #82).

I added a test to demonstrate the issue, but fixing will take a little longer as I'm not at all familiar with the code. I might have some time tomorrow though, and will hopefully fix things.

Hope you're well,

James

### DIFF
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -1,7 +1,8 @@
 (ns cljfmt.core-test
   (:require [#?@(:clj (clojure.test :refer)
                  :cljs (cljs.test :refer-macros)) [deftest testing is]]
-            [cljfmt.core :refer [reformat-string]]))
+            [cljfmt.core :refer [reformat-string]]
+            [clojure.string :as str]))
 
 (deftest test-indent
   (testing "list indentation"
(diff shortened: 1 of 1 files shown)


cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -105,6 +106,23 @@
...
+                (str/join "\n")
+                reformat-string
+                str/split-lines)
+           ["(defn foo"
+            "  []"
+            "  (let [x 1]"
+            "    ;; test1"
+            "    )"
+            "  ;; test2"
+            "  )"]))
```
jcf: I've deviated from the existing format of tests because I really struggle to see what's indented etc. when everything's on a single line with newlines dotted around.

I can pack things up in the same style if you prefer.

jcf:
And as expected the test fails! It's not often I want a build to fail. :smile:

```
FAIL in (test-indent) (core_test.cljc:109)
comment before ending bracket
expected: (= (->> ["(defn foo" "  []" "  (let [x 1]" "    ;; test1" "    )" "  ;; test2" "  )"] (str/join "\n") reformat-string str/split-lines) ["(defn foo" "  []" "  (let [x 1]" "    ;; test1" "    )" "  ;; test2" "  )"])
  actual: (not (= ["(defn foo" "  []" "  (let [x 1]" "    ;; test1" ")" "  ;; test2" "  )"] ["(defn foo" "  []" "  (let [x 1]" "    ;; test1" "    )" "  ;; test2" "  )"]))
```

cljfmt/test/cljfmt/core_test.cljc
weavejester: That seems fine. I was using `\n` for brevity in most cases.

weavejester:
Thanks, closing the PR.

PR #95, opened 2017-04-14
Ordering ns :require

rodnaph:
Hey - I've added the ability to do a basic sort on `ns` `:require` lists.  This is my first time contributing to this library, and first time using `rewrite-clj` so it's pretty rough, I could use some feedback - thanks.


weavejester:
Hi, sorry for getting back to you so late. The notification for this PR seems to have passed me by. Can you add a test for ordering when the requires are on their own lines? We want to make sure it doesn't mess up any line breaks. Can you also change the commit message to:

    Add ordering of ns requires

weavejester:
Closed by 23daaf00

PR #96, opened 2017-06-30
Make indent settings available at runtime

MichaelBlume:
vim-cljfmt uses a REPL connection to call reformat-string and reformat vim
buffers on command. Unfortunately, it doesn't respect indent options specified
in project.clj/profiles.clj. As mentioned in venantius/vim-cljfmt#24 the author
is reluctant to parse these files manually.

By injecting indent settings into cljfmt.core at REPL startup, we can enable
vim-cljfmt to easily respect these settings.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -22,6 +22,9 @@
           :refer [append-space skip whitespace-or-comment?]])
         (:require-macros [cljfmt.core :refer [read-resource]])]))
 
+(def project-settings-store (atom {}))
+(defn project-settings [] @project-settings-store)
+
 #?(:clj (def read-resource* (comp read-string slurp io/resource)))
 #?(:clj (defmacro read-resource [path] `'~(read-resource* path)))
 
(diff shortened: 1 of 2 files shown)


weavejester:
I don't want to add any code specific to a certain editor integration to the cljfmt project. Ideally this should be placed in a separate plugin. For example, the vim-cljfmt project could include a Leiningen plugin that both adds the cljfmt dependency and the necessary injects to the project map via middleware.

MichaelBlume:
So I used vim-cljfmt as a motivating reference in the commit message, but this isn't specific to any particular editor integration. This allows any integration to cljfmt to get the project settings at runtime.

MichaelBlume:
I haven't looked into the emacs plugin, but I could easily see it using this integration point with no modification.

weavejester:
Sure, but I don't think that should be cljfmt's responsibility. Neither cljfmt or lein-cljfmt require this functionality to work, so this functionality should be placed into a separate plugin. Perhaps a lein-vim-cljfmt plugin, or perhaps a lein-cljfmt-integration plugin for a more generic system.

MichaelBlume:
Sure, I can do that

PR #104, opened 2017-10-05
Parallel fix and check

JoelSanchez:
I noticed that cljfmt was not using all of my cores. Hence this PR.
`time` with latest version (0.5.7):
```
real    59,88s
user    65,32s
sys 0,69s
```
`time` with this PR:
```
real    37,71s
user    104,55s
sys 0,74s
```

There are very likely things to improve in this PR, but at least it got slightly more efficient!

### DIFF
--- a/lein-cljfmt/project.clj
+++ b/lein-cljfmt/project.clj
@@ -7,4 +7,5 @@
   :eval-in-leiningen true
   :dependencies [[cljfmt "0.5.7"]
                  [meta-merge "1.0.0"]
+                 [org.clojure/core.async "0.3.443"]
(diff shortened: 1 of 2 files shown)


weavejester:
Thanks for the patch! Could you explain why you chose core.async over using reducers or pmap?

JoelSanchez:
IMO pmap gives you much less control, and I chose core.async over reducers
because I'm more familiar with it :)

rborer:
Any chance to see this feature land in a release? I'm quite interested in it since I have a quite big project that would definitely benefit from this performance improvement 😄 .

Issue #108, opened 2017-12-22
Namespaces not preventing collisions with built-in formatters


MatthewDarling:
Given [identical formatting definitions](https://github.com/MatthewDarling/cljfmt-defrecord-formatting-example/blob/master/project.clj#L8-L9) of `[[:block 1] [:inner 1]]`, a form that [shares a name with `clojure.core` will indent differently](https://github.com/MatthewDarling/cljfmt-defrecord-formatting-example/blob/master/src/core.clj):

```clojure
(s/defrecord Complicated
             [attr1 attr2 attr3 attr4
              attr5 attr6 attr7 attr8]
  component/Lifecycle
  (start [component]
    component)
  (stop [component]
    component))

(defcomponent empty-component
  [_ owner _]
  (:mixins [])

  (init-state [_]
    {})

  (render-state [_ state]
    (html [:div])))
``` 

I've found that using `^:replace` will get them to indent the same. However, I sort of expected that a namespaced symbol like `s/defrecord` could have different indentation from `clojure.core/defrecord`. Not even at the level of "resolve the `s` alias and check the indentation rules for its fully qualified name", just like, look up the namespaced symbol in the configuration and use whatever is there.

One downside of respecting the namespace would be potentially having configuration like:

```clojure
{s/defrecord [[:block 1] [:inner 1]]
 schema.core/defrecord [[:block 1] [:inner 1]]
 schema/defrecord [[:block 1] [:inner 1]]}
```

But consistent namespace aliases are good practice, so maybe the extra configuration will cause the user to fix the inconsistency :)

MatthewDarling:
As an aside, I think the behaviour of `^:replace` could be clarified a bit in the readme.

Here's a rough attempt:

```markdown
As with Leiningen profiles, you can add metadata hints. If you want 
to override any existing indents, the `:replace` hint will allow that:

    ```clojure
    :cljfmt {:indents ^:replace {cond [[:inner 0]]}}
    ```

If you want to have a clean slate for defining indents, the replace hint 
combined with a greedy regular expression will allow that:

    ```clojure
    :cljfmt {:indents ^:replace {#".*" [[:inner 0]]}}
    ```

From there, you could define indents one at a time to achieve the desired behaviour.
```

----------

[...]

Issue #106, opened 2017-10-24
Namespaced keywords are not supported


kennyjwilli:
Parsing the below string using rewrite-cljs or rewrite-clj works as expected:
```clojure
(rewrite-clj.node/string
  (rewrite-clj.parser/parse-string-all "(ns myns.core)\n\n{::foo-bar (do (map)\n               (foo))}"))
=> "(ns myns.core)\n\n{::foo-bar (do (map)\n               (foo))}"
```

Using cljfmt throws this exception:
```clojure
(cljfmt.core/reformat-string "(ns myns.core)\n\n{::foo-bar (do (map)\n               (foo))}")
repl:13
throw e__9355__auto__;
^
Error: Namespaced keywords not supported !
    at rewrite_clj.node.keyword.KeywordNode.rewrite_clj$node$protocols$Node$sexpr$arity$1 (/home/kenny/compute_software/cljfmt-cli/.cljs_node_repl/rewrite_clj/node/keyword.js:149:8)
    at rewrite_clj$node$protocols$sexpr (/home/kenny/compute_software/cljfmt-cli/.cljs_node_repl/rewrite_clj/node/protocols.js:62:10)
    at rewrite_clj$zip$base$sexpr (/home/kenny/compute_software/cljfmt-cli/.cljs_node_repl/rewrite_clj/zip/base.js:57:31)
    at cljfmt$core$token_value (/home/kenny/compute_software/cljfmt-cli/.cljs_node_repl/cljfmt/core.js:318:30)
    at cljfmt$core$form_symbol (/home/kenny/compute_software/cljfmt-cli/.cljs_node_repl/cljfmt/core.js:332:71)
    at /home/kenny/compute_software/cljfmt-cli/.cljs_node_repl/cljfmt/core.js:339:131
    at cljfmt$core$inner_indent (/home/kenny/compute_software/cljfmt-cli/.cljs_node_repl/cljfmt/core.js:350:3)
    at cljs.core._add_method.call.vec__15752 (/home/kenny/compute_software/cljfmt-cli/.cljs_node_repl/cljfmt/core.js:441:33)
    at Function.cljs.core.some_fn.cljs$core$IFn$_invoke$arity$2.cljs$core$sp2__1 [as cljs$core$IFn$_invoke$arity$1] (/home/kenny/compute_software/cljfmt-cli/out/.cljs_node_repl/cljs/core.js:16820:85)
[...]

PR #109, opened 2017-12-23
Custom indentation per-namespace

MatthewDarling:
This PR implements #108, supporting indent rules that apply only to a namespaced symbol. When a namespaced symbol is checked for a matching indent rule, it will check first for a namespaced version, then fall back to the old behaviour if none is found.

## Proof that it works

You can see the [difference in behaviour in this example repo](https://github.com/MatthewDarling/cljfmt-defrecord-formatting-example/compare/namespace-support) - it finds that the `s` alias resolves to `schema.core`, and then uses the `schema.core/defrecord` indentation rules.

## Assorted thoughts about the changes

  * I implemented `find-all` as a way to get a seq of matching nodes in a zipper. This is the sort of thing that I assume someone else has already written, but I can never find it when I need it. Here, it's used to find all the `:as` notices within a `:require` block.
  * `indent-order` now has this ordering: namespaced symbols, all other symbols, regex patterns
  * Adding `alias-map` as an argument to a bunch of functions makes the diff look ugly. Public functions received an extra arity to accept it, or will use a default value. Private functions now require it as a mandatory argument. Likely there's room for improvement here though.
  * I'm not sure the updated documentation belongs where it is. Maybe a separate "advanced configuration settings" section might be good? I tried to be really specific about the niche situations that this functionality is useful. Still, I'm concerned users might reach for this stuff when they don't need it, because it's mixed in with the more basic info.

## CLJS issues

There's already an [open issue about `rewrite-cljs` not fully tracking `rewrite-clj`](https://github.com/weavejester/cljfmt/issues/106), and it bit me here. All the functions used to find aliases inside an `ns` block require some of the missing behaviour.

Could be worked around by using `tools.analyzer` in CLJS for this specific task, or a hacky `map`/`filter`/etc combo, but... it's probably just better for someone to update `rewrite-cljs`.

To get around the issue for now, `format-string` checks its options map for the key `:alias-map`. That way users can manually fill out any aliases they need. Hopefully this workaround isn't necessary for long. Maybe the README should say "help wanted to update `rewrite-cljs`" in the relevant section?

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,13 +108,48 @@ You can also configure the behavior of cljfmt:
   a map of var symbols to indentation rules, i.e. `{symbol [& rules]}`.
   See the next section for a detailed explanation.
 
-As with Leiningen profiles, you can add metadata hints. If you want to
-override all existing indents, instead of just supplying new indents
-that are merged with the defaults, you can use the `:replace` hint:
+  Unqualified symbols in the indents map will apply to any symbol with a
+  matching "name" - so `foo` would apply to both `org.me/foo` and
+  `com.them/foo`. If you want finer-grained control, you can use a fully
+  qualified symbol in the indents map to configure indentation that
+  applies only to `org.me/foo`:
 
-```clojure
-:cljfmt {:indents ^:replace {#".*" [[:inner 0]]}}
-```
+  ```clojure
+  :cljfmt {:indents {org.me/foo [[:inner 0]]}}
+  ```
+
+  Configured this way, `org.me/foo` will indent differently from
+  `com.them/foo`.
+
+  Note that `cljfmt` currently doesn't resolve symbols brought into a
+  namespace using `:refer` or `:use` - they can only be controlled by an
+  unqualified indent rule.
+
+  As with Leiningen profiles, you can add metadata hints. If you want to
+  override all existing indents, instead of just supplying new indents
+  that are merged with the defaults, you can use the `:replace` hint:
+
+  ```clojure
+  :cljfmt {:indents ^:replace {#".*" [[:inner 0]]}}
+  ```
+
+* `:alias-map` -
+  a map of namespace alias strings to fully qualified namespace
+  names. This option is unnecessary in almost all cases, because
+  `cljfmt` can compute the alias map from an `ns`
+  declaration.
+
+  However, it can't do that when used as a CLJS library,
+  or when indenting something with no `ns` declaration like an EDN
+  file. Even in those situations, you only need this option when using
+  indentation rules that rely on the fully qualified symbol name.
+
+  If you definitely need to configure this, it should look like this:
+
+  ```clojure
+  :cljfmt {:indents {org.me/foo [[:inner 0]]}
+           :alias-map {"me" "org.me"}}
+  ```
 
 
 ### Indentation rules
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -35,6 +35,14 @@
   #?(:clj  (fn [^String a ^String b] (.contains a b))
      :cljs str/includes?))
 
+(defn- find-all [zloc p?]
+  (loop [matches []
+         zloc zloc]
+    (if-let [zloc (z/find-next zloc zip/next p?)]
+      (recur (conj matches zloc)
+             (zip/next zloc))
+      matches)))
+
 (defn- edit-all [zloc p? f]
   (loop [zloc (if (p? zloc) (f zloc) zloc)]
     (if-let [zloc (z/find-next zloc zip/next p?)]
@@ -194,15 +202,23 @@
   (and (reader-macro? zloc) (#{"?" "?@"} (-> zloc z/down token-value str))))
 
 (defn- form-symbol [zloc]
-  (-> zloc z/leftmost token-value remove-namespace))
+  (-> zloc z/leftmost token-value))
 
 (defn- index-matches-top-argument? [zloc depth idx]
   (and (> depth 0)
        (= idx (index-of (nth (iterate z/up zloc) (dec depth))))))
 
-(defn- inner-indent [zloc key depth idx]
+(defn- fully-qualify-symbol [possible-sym alias-map]
+  (if-let [ns-string (and (symbol? possible-sym)
+                          (namespace possible-sym))]
+    (symbol (get alias-map ns-string ns-string)
+            (name possible-sym))
+    possible-sym))
+
+(defn- inner-indent [zloc key depth idx alias-map]
   (let [top (nth (iterate z/up zloc) depth)]
-    (if (and (indent-matches? key (form-symbol top))
+    (if (and (or (indent-matches? key (fully-qualify-symbol (form-symbol top) alias-map))
+                 (indent-matches? key (remove-namespace (form-symbol top))))
              (or (nil? idx) (index-matches-top-argument? zloc depth idx)))
       (let [zup (z/up zloc)]
         (+ (margin zup) (indent-width zup))))))
@@ -219,11 +235,12 @@
       (or (zlinebreak? zloc) (comment? zloc)))
     true))
 
-(defn- block-indent [zloc key idx]
-  (if (indent-matches? key (form-symbol zloc))
+(defn- block-indent [zloc key idx alias-map]
+  (if (or (indent-matches? key (fully-qualify-symbol (form-symbol zloc) alias-map))
+          (indent-matches? key (remove-namespace (form-symbol zloc))))
     (if (and (some-> zloc (nth-form (inc idx)) first-form-in-line?)
              (> (index-of zloc) idx))
-      (inner-indent zloc key 0 nil)
+      (inner-indent zloc key 0 nil alias-map)
       (list-indent zloc))))
 
 (def default-indents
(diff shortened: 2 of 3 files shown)


weavejester:
This looks good. I also assumed there must also be some sort of `find-all` function already written, but if there is I couldn't find it either. I think waiting until rewrite-cljs is updated is the right call, and while passing `alias-map` around does increase the number of arguments, I can't think of a better way to do it.

From my reading of the code, it looks like this doesn't support the nested namespace syntax?

MatthewDarling:
Oh, good catch on nesting. I'm sure it doesn't work for that. I'll take a look later this week.

Also, I think it might not work as-is for namespaces that use reader conditionals in the `ns` clause. I'll double check on that, too.

MatthewDarling:
Added support for reader conditionals by introducing a function `top-level-containing-form`, rather than just using `z/up`. The one call to `z/up` was finding the node for the conditional and then breaking things. Now `alias-map-for-form` works on the `cljfmt.core` itself :)

I also got support for nested `require` by checking if the parent is the `:require` symbol - if not, it's taken to be a parent namespace, and they're joined together.

So this is supported:

```clojure
(ns cljfmt.nested
  (:require [clojure
             [zip :as zip]
             [string :as str]
             [set :as set]]))
```

and returns `{"zip" "clojure.zip" "str" "clojure.string" "set" "clojure.set"}`.

----------

However, it only goes up one level, so it would break in a case like this:

```clojure
(ns cljfmt.nested
  (:require [clojure
             [zip :as zip]
             [string :as str]
             [set :as set]
             [tools
              [reader :as reader]]]))
```

But that produces a compiler error saying "no value supplied for key: true". I guess if it's invalid syntax, `cljfmt` wouldn't need to support it... Are you aware of multiple levels of nesting being a possibility?

MatthewDarling:
By the way, with reader conditional support, using `find-all` from the first `:require` symbol happens to find all the `:as` nodes in both the CLJ `:require` and the CLJS `:require`. Which is nice that it's no extra work.

However, it's possible (I think?) to use the same alias for different namespaces, depending on environment:

```clojure
(ns cljfmt.core
  #?@(:clj
      [(:require
        [clojure.tools.reader :as reader])]
      :cljs
      [(:require
        [cljs.reader :as reader])]))
```

And that would produce unexpected behaviour - it might indent using `cljs.reader` rules, or it might indent using `clojure.tools.reader` rules. It seemed consistent across multiple runs (always using the `cljs.reader` rules for me), but it's unpredictable.

I've experimented a little bit with a way to fix this, but it makes the code more complicated. The first step is to change the alias map from:

```clojure
{"reader" "clojure.tools.reader"}
```

to

```clojure
{"reader" #{"cljs.reader" "clojure.tools.reader"}}
```

And then indent symbols as if `cljs.reader` and `clojure.tools.reader` are the same. This prevents `cljs.reader/foo` and `clojure.tools.reader/foo` from having different indentation rules, but would have more predictable behaviour.

Anyway, I can push those changes to another branch if you want to take a look.

arrdem:
AFAIK `[clojure [tools [reader ...]]]` is not supported syntax. I was reading the implementation of `(ns)` last night for unrelated reasons.

weavejester:
I'd like to take a look. It might be possible to accept this PR as is, then work on compatibility with conditionals in another PR.

By "nested namespace syntax" I meant the prefix syntax, e.g.

```clojure
(require '[clojure.java [io :as io]])
```

I might be slow merging this, as I'm on vacation until the new year :)

weavejester:
Thanks for waiting. Does this PR support prefix syntax in namespaces? Also, are you happy with the current patch or do you plan on adding more changes?

MatthewDarling:
Yep, prefix syntax is supported. Added [in this commit](https://github.com/weavejester/cljfmt/pull/109/commits/2d764ba633b5478c1d827a48215a8f672239d7e5).

I'm not happy with the handling of aliases within reader conditionals. It's only a problem if the same alias is used for different namespaces, but that seems like a nice pattern if you have platform-specific libraries with compatible APIs.

But, I don't have a specific solution in mind either.

Do you have any ideas for a better approach?

weavejester:
If they're compatible APIs, wouldn't we want to indent them the same anyway? Regardless, it still seems an improvement over the current system.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -306,11 +328,68 @@
       (cond-> (:insert-missing-whitespace? opts true)
         insert-missing-whitespace)
       (cond-> (:indentation? opts true)
-        (reindent (:indents opts default-indents)))
+        (reindent (:indents opts default-indents)
+                  (:alias-map opts {})))
       (cond-> (:remove-trailing-whitespace? opts true)
         remove-trailing-whitespace)))
 
+(defn- top-level-containing-form
```
weavejester: I think this could be written:

```clojure
(defn- top-level-form [zloc]
  (->> zloc
       (iterate z/up)
       (take-while (complement root?))
       last))
```

Without losing any information. As this isn't a public API, we don't need a docstring if the source code and function name makes the purpose of the function clear. To my mind, `(top-level-form zloc)` is self-explanatory.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -306,11 +328,68 @@
...
+  function finds the start of the subtree that contains `node`."
+  [node]
+  (->> node
+       (iterate z/up)
+       (take-while (complement root?))
+       last))
+
+#?(:clj
+   (defn- require-node?
+     [node]
```
weavejester: Can you put the argument list on the same line as the function name, as long as there's room to do so?

Also I tent to use `zloc` elsewhere in the source code to denote an arbitrary zipper location, rather than node. Consistent naming conventions should make the code easier to read.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -306,11 +328,68 @@
...
+  zipper for that namespace will have all those forms as children. This
+  function finds the start of the subtree that contains `node`."
+  [node]
+  (->> node
+       (iterate z/up)
+       (take-while (complement root?))
+       last))
+
+#?(:clj
+   (defn- require-node?
```
weavejester: Let's call this: `ns-require-form?`

cljfmt/src/cljfmt/core.cljc
```diff
@@ -306,11 +328,68 @@
...
+     [node]
+     (and (some-> node
+                  top-level-containing-form
+                  z/child-sexprs
+                  first
+                  (= 'ns))
+          (some-> node z/child-sexprs first (= :require)))))
+
+#?(:clj
+   (defn- as-node?
```
weavejester: Let's call this: `as-keyword?`

```clojure
(defn- as-keyword? [zloc]
  (and (= :token (z/tag zloc))
       (= :as (z/sexpr zloc))))

File: cljfmt/src/cljfmt/core.cljc [hunk 5]
```diff
@@ -306,11 +328,68 @@
...
+                              (first (z/child-sexprs grandparent-node)))]
+       (when (and (symbol? alias) (symbol? current-namespace))
+         {(str alias) (if parent-namespace
+                        (format "%s.%s" parent-namespace current-namespace)
+                        (str current-namespace))}))))
+
+#?(:clj
+   (defn- alias-map-for-form
+     [form]
+     (when-let [require-zloc (-> form
```
weavejester: The `when-let` clause could be all on one line.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -306,11 +328,68 @@
...
+#?(:clj
+   (defn- as-node?
+     [node]
+     (and (= :token (z/tag node))
+          (= :as (z/sexpr node)))))
+
+#?(:clj
+   (defn- as-zloc->alias-mapping
+     [as-zloc]
+     (let [alias (some-> as-zloc z/right z/sexpr)
```
weavejester: Can you align the let clauses for clarity?

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -64,7 +64,15 @@
     (is (= (reformat-string "(t/defn foo [x]\n(+ x 1))")
            "(t/defn foo [x]\n  (+ x 1))"))
     (is (= (reformat-string "(t/defrecord Foo [x]\nCloseable\n(close [_]\n(prn x)))")
-           "(t/defrecord Foo [x]\n  Closeable\n  (close [_]\n    (prn x)))")))
+           "(t/defrecord Foo [x]\n  Closeable\n  (close [_]\n    (prn x)))"))
+    (is (= (reformat-string "(ns example\n(:require [thing.core :as t]))\n\n(t/defn foo [x]\n(+ x 1))\n\n(defn foo [x]\n(+ x 1))"
```
weavejester: Can you make sure that the lines do not exceed 80 characters?

weavejester:
I just have a few minor formatting tweaks, but other than that I think the code looks good.

Once the formatting is done, we can squash any extraneous commits down so that the code history will be clean once merged.

cljfmt/test/cljfmt/core_test.cljc
MatthewDarling: Cribbed from [this PR](https://github.com/weavejester/cljfmt/pull/88/files#r92662161) for how to do that, seems like a good fix for the issue :)

README.md
```diff
@@ -108,13 +108,46 @@ You can also configure the behavior of cljfmt:
   a map of var symbols to indentation rules, i.e. `{symbol [& rules]}`.
   See the next section for a detailed explanation.
 
-As with Leiningen profiles, you can add metadata hints. If you want to
+  Unqualified symbols in the indents map will apply to any symbol with a
```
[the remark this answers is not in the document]
MatthewDarling: The two space indent is needed to keep the extra paragraphs as part of the current list item. Compare the rendered result, using a numbered list so it's easier to tell:

## Existing format

1. List item 1
1. List item 2

More details about list item 2

1. List item 3

## New format

1. List item 1
1. List item 2

    More details about list item 2

1. List item 3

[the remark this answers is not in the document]
weavejester: In which case, can you indent each line of the paragraph? I just need the formatting to be consistent.

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -64,7 +65,55 @@
...
+           "(t/defrecord Foo [x]\n  Closeable\n  (close [_]\n    (prn x)))"))
+    (is (= (-> ["(ns example"
+                "(:require [thing.core :as t]))"
+                ""
+                "(t/defn foo [x]"
+                "(+ x 1))"
+                ""
+                "(defn foo [x]"
+                "(+ x 1))"]
+               (->> (str/join "\n"))
```
weavejester: It might look a little better if you used `str`, e.g.

```clojure
(str "(ns example\n"
     "(:require [thing.core :as t]))\n\n"
     "(t/defn foo [x]\n"
     "(+ x 1))\n\n"
     "(defn foo [x]\n"
     "(+ x 1))")
```

MatthewDarling:
Thanks for the feedback and review :) Think I'm caught up with the comments so far, let me know if you spot anything else.

weavejester:
Thanks! This looks fine. Can you squash down the commits so there's a clean history to merge?

weavejester:
Thanks. Can you remove the outdated comments from the commit message and change the commit message to something like:

    Add namespace-specific indentation

MatthewDarling:
Let me know if the new squash looks okay :)

weavejester:
Thanks! That looks all fine!

PR #113, opened 2018-01-26
Add ns :require sorting support

kipz:
- Add a new option: `:sort-ns-requires?` which defaults to `false` for now
- Only sorts `:require` vectors (ignores symbols, lists etc)

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -297,6 +297,34 @@
 (defn remove-trailing-whitespace [form]
   (transform form edit-all trailing-whitespace? zip/remove))
 
+
+(defn sort-requires
+  [require-zloc]
+  (z/replace
+   require-zloc
+   (n/replace-children
+    (z/node require-zloc)
+    (cons
+     (n/keyword-node :require)
+     ; because we want to keep the current whitespace and only reorder the requires
+     (loop [unsorted-requires (->> require-zloc z/node n/children rest)
+            sorted-requires (->> unsorted-requires (filter #(= :vector (n/tag %))) (sort #(compare (n/sexpr %1) (n/sexpr %2))))
+            new-children []]
+       (if (empty? unsorted-requires)
+         new-children
+         (let [unsorted-require (first unsorted-requires)]
+           (if (= :vector (n/tag unsorted-require))
+             (recur (rest unsorted-requires) (rest sorted-requires) (conj new-children (first sorted-requires)))
+             (recur (rest unsorted-requires) sorted-requires (conj new-children (first unsorted-requires)))))))))))
+
+(defn require-node? [zloc]
+  (and
+   (= :require (z/sexpr (z/down zloc)))
+   (= (symbol "ns") (z/sexpr (z/next (z/up zloc))))))
+
+(defn sort-ns-requires [form]
+  (transform form edit-all require-node? sort-requires))
+
 (defn reformat-form [form & [{:as opts}]]
   (-> form
       (cond-> (:remove-consecutive-blank-lines? opts true)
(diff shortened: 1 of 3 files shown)


weavejester:
Thanks for the PR.

I'll have time for a more thorough review later, but my initial thoughts is that your function is too large, and you're not making use of existing functions so you're duplicating functionality.

You're also using your comments to explain *what* rather than *why*; if you're doing that, it's a strong indication that you need to split up your function instead. For example, you've written:

```clojure
; check if this is a require node
(if-let [req (z/find
              (z/right node)
              z/next
              (fn [[{:keys [children]}]]
                (= :require (-> children first :k))))]
```

But a better way would be to write:

```clojure
(if (require-node? node)
```

See how it conveys the same information, but without the need for a comment? Comments should be used sparingly, to explain *why* in the cases where it's not obvious. If the *what* isn't obvious, then more functions are required.

However, I don't think you need to use a loop at all. Instead, use `transform` and `edit-all`:

```clojure
(defn sort-ns-requires [form]
  (transform form edit-all ns-require? sort-namespaces))
```

This will traverse the form, looking for a zipper loc that adheres to `ns-require?` and then applies `sort-namespaces` to transform it. You'll need to write those two functions, of course.

I'll be able to give you more detailed suggestions later on.

kipz:
Thanks @weavejester for the suggestions - I missed some useful utility functions! I've made some the of the changes you suggest, though I think the `sort-requires` function needs some more work - perhaps there are some more handy things in there to help?

cljfmt/src/cljfmt/core.cljc
```diff
@@ -297,6 +297,31 @@
 (defn remove-trailing-whitespace [form]
   (transform form edit-all trailing-whitespace? zip/remove))
 
+
+(defn sort-requires
+  [require-zloc]
+  (z/replace
```
weavejester: Can you explain what you're doing with this function? Also how does it handle changing the indent level?

cljfmt/src/cljfmt/core.cljc
```diff
@@ -297,6 +297,31 @@
...
+            new-children []]
+       (if (empty? unsorted-requires)
+         new-children
+         (let [unsorted-require (first unsorted-requires)]
+           (if (= :vector (n/tag unsorted-require))
+             (recur (rest unsorted-requires) (rest sorted-requires) (conj new-children (first sorted-requires)))
+             (recur (rest unsorted-requires) sorted-requires (conj new-children (first unsorted-requires)))))))))))
+
+(defn require-node? [zloc]
+   (= :require (z/sexpr (z/down zloc))))
```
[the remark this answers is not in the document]
kipz: Good catch. I'll add another test for that.

cljfmt/src/cljfmt/core.cljc
kipz: This bit is a little tricky. First it collect all the vectors within the `(:require..)`, sorts them, then effectively plays them back in sorted order in the loop, replacing only the vector nodes. This leaves all the original whitepspace nodes in place.

weavejester: That's an interesting approach. It would mean that comments wouldn't follow their associated clause, but comments are rare in the `ns` declaration anyway.

You seem to be looking for vectors, but namespaces in `require` can just be symbols. Perhaps instead look for non-whitespace tokens?

Maybe you want something more like:

```clojure
(let [nodes  (partition-by whitespace? children)
      spaces (take-nth 2 nodes)
      forms  (take-nth 2 (rest nodes))]
  (->> forms (sort-by sexpr) (interleave spaces) (apply concat)))
```

Or even factor out the unravelling:

```clojure
(defn unravel [n coll]
  (map #(take-nth n (drop % coll)) (range n)))

;; ...

(let [[spaces forms] (->> children (partition-by whitespace?) (unravel 2))]
  (->> forms (sort-by sexpr) (interleave spaces) (apply concat)))
```

Obviously that's pseudo-ish code, but should give you a rough idea of what I mean.

kipz: I think it'd be cool if it worked with comments too, so I'll take a look at that. We can't rely on the same hard-coded unraveling then though.

weavejester:
Closed by 23daaf00

Issue #118, opened 2018-04-28
New lines disappear when `\n\n\n ` between forms


snoe:
When there are at least three newlines followed by a space between two forms, calling `reformat-string` will delete the newlines:
`(cljfmt/reformat-string "a\n\n\n b") => "a b"`

expected:
`(cljfmt/reformat-string "a\n\n\n b") => "a\n\nb"`

ProjectFrank:
was this issue sufficiently addressed in the PR that references it?

weavejester:
Yes, looks like this is fixed.

PR #119, opened 2018-04-28
When encountering greater than 2 consecutive lines

snoe:
completely wipeout whitespace and newlines before inserting 2 consecutive lines

Addresses #118

Since `remove` moves right to left the `(zip/replace (n/newlines 2))` was getting clobbered by `remove-whitespace-and-newlines`. This slipped through because none of the existing tests had whitespace nodes after `consecutive-blank-line?` was true

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -108,7 +108,12 @@
     zloc))
 
 (defn- replace-consecutive-blank-lines [zloc]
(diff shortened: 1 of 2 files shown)


weavejester:
Thanks for the fix! Can you adjust your commit message so it adheres to the [seven rules](https://chris.beams.io/posts/git-commit/#seven-rules), and adjust the line you changed so that it comes under 80 characters.

snoe:
@weavejester how's that?

weavejester:
Almost perfect - just remove the "for", so "Fix consecutive lines followed by whitespace".

PR #123, opened 2018-05-26
Extract leiningen.cljfmt functionality to the core project

vemv:
AFAICT, namespaces in the form `leiningen.*` are somewhat magical and cannot be required outside the context of a Lein plugin.

This makes it difficult to programatically use cljfmt, e.g. I want to build a custom formatting facility.

Last time I encountered this problem I had to copy/paste and adapt the whole `leiningen.cljfmt` which obviously is far from ideal.

This PR moves code around, leaving it untouched. Only IMain/Main/*main* are introduced, so the new ns doesn't depend on `leiningen.core.main`

I tried out these changes locally, they work!

Cheers - Victor


weavejester:
Thanks for the PR! I like the idea of moving this to a namespace that doesn't use Leiningen, but I don't think we should leave the Leiningen idioms like `info`, `abort` etc. in place if we do.

Instead, I think it might be better to have a `cljfmt.main` namespace, with a corresponding `-main` function. We can expose `fix`, `check` and any other function that seems useful as part of the public API, but the Leiningen compatibility protocol should be removed. Instead, I think that should be changed to use `*out*`, `*err*` and `System/exit` directly.

The only thing I think we'd lose from Leiningen is `LEIN_SILENT`, which we can implement by just rebinding `*err*` and `*out*` when `leiningen.core.main/*info*` is falsey.

vemv:
Feedback implemented! I hope my understanding was correct.

Tried out the code again with:

```
LEIN_SILENT=1 lein cljfmt check
lein cljfmt check
LEIN_SILENT=1 lein cljfmt fix
lein cljfmt fix
```

So now the diff with the original code is:

* `defn abort ...`
* `defn warn ...`
* A new `-main` function
* `leiningen.cljfmt` uses said main plus a 'sink' to possibly swallow `*out*`/`*err*`

weavejester:
Thanks for the update. Most of the functions in `cljfmt.main` can be private, as we don't want to expose them as a public API. We also need to replace `project` with command-line options instead.

weavejester:
So, just to clarify: `fix` and `check` should take a map of options, and `-main` should use [tools.cli](https://github.com/clojure/tools.cli) to parse those options from the command line.

The Leiningen plugin will take the relevant options from the project map, and pass them directly to `fix` and `check`.

vemv:
Hey James,

thanks for the feedback! Here's a tentative take on it:

https://github.com/vemv/cljfmt/commit/5559e0ac8e029a8734e77989cce9bb5165e0a513

Let me know if it's well-guided.

The only current flaw I'm aware of is that `:indents` and `:alias-map` are rather complex types, so I don't know if we should realistically expect people will pass that from the cli.

To be clear I'd expect cljfmt to have three use cases now:

* From lein, as always
* From clojure libraries, with full options just like lein-cljmt
* From cli, _perhaps_ without the :indents/:alias-map options.

weavejester:
Thanks for the update!

That looks good, but don't use `binding` and `*options*` - pass the options to the functions as an argument. Essentially just do it the same as `project`, except we can get rid of the Leiningen-specific stuff.

Complex arguments can be ommitted from `-main`, since we can always use `fix` and `check` directly. However, it might be a good idea to make the `:indents` and `:alias-map` options take a file that we just `slurp` and `edn/read-string` from.

vemv:
Understood, thanks for reviewing!

> However, it might be a good idea 

Shall I implement it?

weavejester:
Sure. Hopefully it should be as easy as `(edn/read-string (slurp indents-file))`.

vemv:
`--no-ansi` is OK, it's `--no-ansi?` the one that would be confusing.

For being able to use `--no-ansi`, I'd have to perform the mentioned rename through the codebase do Clojure code and CLI options match in name.

weavejester:
Ah, I see. I think you can set the key name explicitly in tools.cli by adding a `:id` key to the options definition. So `:id :ansi?`might work.

cljfmt/src/cljfmt/diff.clj
```diff
@@ -9,13 +10,18 @@
 (defn unlines [ss]
   (str/join "\n" ss))
 
+(defn to-absolute-path [filename]
```
vemv: Not sure if this defn should go here. Else I'd suggest a tiny `utils` namespace.

cljfmt/src/cljfmt/diff.clj
```diff
@@ -9,13 +10,18 @@
...
+       .getCanonicalPath))
+
 (defn unified-diff
   ([filename original revised]
    (unified-diff filename original revised 3))
   ([filename original revised context]
    (unlines (DiffUtils/generateUnifiedDiff
-             (str (io/file "a" filename))
-             (str (io/file "b" filename))
+             (->> filename to-absolute-path (str "a"))
```
vemv: Example cases fixed with this:

* `lein cljfmt check ../../some-dir`
* `lein cljfmt check ~/some-dir`

cljfmt/src/cljfmt/diff.clj
weavejester: Do we need it in another namespace? If not, just make it private. All other functions we don't want as part of the public API need to be private as well, like `lines` and `unlines`.

cljfmt/src/cljfmt/main.clj
```diff
@@ -0,0 +1,189 @@
...
+    :parse-fn re-pattern]
+   [nil "--indents INDENTS_PATH" "The path to a file containing the edn data for indentation rules"
+    :parse-fn cli-file-reader]
+   [nil "--alias-map ALIAS_MAP_PATH" "The path to a file containing the edn data for the alias map"
+    :parse-fn cli-file-reader]
+   [nil "--project-root PROJECT_ROOT" "The directory to be considered the project root"
+    :default (:project-root default-config)]
+   [nil "--[no-]ansi" "Whether to colorize output"
+    :default (:ansi? default-config)
+    :id :ansi]
```
weavejester: Should this be `:ansi?`?

cljfmt/src/cljfmt/main.clj
```diff
@@ -0,0 +1,189 @@
...
+    :id :remove-trailing-whitespace?]
+   [nil "--[no-]insert-missing-whitespace"
+    :default (:insert-missing-whitespace? default-config)
+    :id :insert-missing-whitespace?]
+   [nil "--[no-]remove-consecutive-blank-lines"
+    :default (:remove-consecutive-blank-lines? default-config)
+    :id :remove-consecutive-blank-lines?]])
+
+(defn -main
+  "Format Clojure source files"
```
weavejester: I don't think we need a docstring on the `-main` function, as it's not strictly part of the public API.

lein-cljfmt/src/leiningen/cljfmt.clj
```diff
@@ -1,150 +1,35 @@
...
   "Format Clojure source files"
-  [project command & args]
-  (case command
-    "check" (apply check project args)
-    "fix"   (apply fix project args)
-    (main/abort "Unknown cljfmt command:" command)))
+  [project command & paths]
+  (let [info? leiningen.core.main/*info*
+        sink (when-not info?
+               (proxy [java.io.Writer] []
```
weavejester: This can probably be replace with a `with-out-str`.

cljfmt/src/cljfmt/main.clj
```diff
@@ -0,0 +1,189 @@
...
+   :indentation? true
+   :insert-missing-whitespace? true
+   :remove-surrounding-whitespace? true
+   :remove-trailing-whitespace? true
+   :remove-consecutive-blank-lines? true
+   :indents cljfmt/default-indents
+   :alias-map {}})
+
+(defn- abort
+  [& msg]
```
weavejester: Can you put the arguments on the same line as the function name when there's room to do so?

cljfmt/src/cljfmt/main.clj
```diff
@@ -0,0 +1,189 @@
...
+      (warn incorrect "file(s) formatted incorrectly"))
+    (when (and (zero? incorrect) (zero? error))
+      (println "All source files formatted correctly"))))
+
+(defn- merge-counts
+  ([]    zero-counts)
+  ([a]   a)
+  ([a b] (merge-with + a b)))
+
+(defn check
```
weavejester: We probably want a brief docstring here.

cljfmt/src/cljfmt/main.clj
```diff
@@ -0,0 +1,189 @@
...
+  [options paths]
```
weavejester: Ideally `options` should come after `paths`, and we should also allow a one-arity form, e.g. `(check paths)`.

cljfmt/src/cljfmt/main.clj
```diff
@@ -0,0 +1,189 @@
...
+                      (map (fn [status]
+                             (print-file-status options status)
+                             (:counts status))))
+                (completing merge-counts)
+                paths)]
+    (print-final-count counts)
+    (exit counts)))
+
+(defn fix
+  [options paths]
```
weavejester: Same deal as `check`.

vemv:
Ready again! `:ansi` fix is in rebased `Make cljfmt.main able to take CLI options` commit, the rest in `PR Feedback` commit.

The `to-absolute-path` fn is used from `main`, hence my original question.

cljfmt/src/cljfmt/main.clj
```diff
@@ -0,0 +1,192 @@
...
+    :default (:insert-missing-whitespace? default-config)
+    :id :insert-missing-whitespace?]
+   [nil "--[no-]remove-consecutive-blank-lines"
+    :default (:remove-consecutive-blank-lines? default-config)
+    :id :remove-consecutive-blank-lines?]])
+
+(defn -main
+  [& args]
+  (let [{:keys [errors options] [command & paths] :arguments} (cli/parse-opts args cli-options)]
+    (let [options (meta-merge default-config options)]
```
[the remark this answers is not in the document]
vemv: 👍 moved back to `lein-cljfmt`

vemv:
Amended! No worries, thanks for reviewing

weavejester:
Thanks for the updates. This looks fine as far as I can tell! Can you squash your commits and I'll merge it in for you.

vemv:
🍻!

The commit history is rather meaningful, it allows one to differentiate the moving-around of code (`Extract leiningen.cljfmt functionality to the core project`) from the further developed functionality (`Make cljfmt.main able to take CLI options `). Conflating them can make debugging harder at a later point.

If you don't particularly agree I'd invite you to use the github merge button, probably it has no downsides?

weavejester:
Keeping some separate commits is fine - I'm mainly concerned with making sure the commits present a clean and understandable history. However, commits like "PR feedback" should be squashed out, as they don't aid future understanding.

The Github "squash-merge" button is somewhat of a misnomer, since it's really "squash-fastforward". It loses the merge commit, which isn't what I want. I'd rather keep useful information, and discard commits that make the history more difficult to understand.

vemv:
> So, commits like "PR feedback" should be squashed out

Disagree but that doesn't matter much for an occasional contribution.

> The Github "squash-merge" button is somewhat of a misnomer

Right!

Squashed, and double-checked that that step was correct.

weavejester:
Keeping some of the commits would have been fine. The main thing is to trim commits that don't aid people's understanding. For example, if there's a commit that fixes some typo or error, then it's a good idea to remove it before merging into Master because it doesn't add any additional explanation about what the feature is about. In the same way, when I rewrite sentences in this comment, I delete them instead of striking them out.

Can you also trim the subject of the commit to 50 characters, and wrap the commit body at 72? I haven't gotten around to adding this to the cljfmt repo yet, but I follow these [contributing guidelines](https://github.com/weavejester/contributing/blob/master/CONTRIBUTING.md).

vemv:
> The main thing is to trim commits that don't aid people's understanding.

The thing is, if the latest commit is a "PR feedback" one, squashing it would mean to either meld it into the previous one (which would break the meaning of the prior commit message), or distribute it evenly with an intrincate rebasing (which I'm used to, but doesn't seem worth the effort for stylistic improvements).

As a rule of thumb, I go by "the last commit of a branch may be called `PR feedback`, as long as it doesn't contain new features/fixes".

Otherwise yeah, "wip" or "fix" are just wrong.

Of course, that's just my POV which I'm just sharing. I respect your stance, especially coming from a reputed author.

> Can you also trim the subject of the commit to 50 characters, and wrap the commit body at 72? I haven't gotten around to adding this to the cljfmt repo yet, but I follow these contributing guidelines.

Honestly I can't come up with something shorter than `Extract leiningen.cljfmt functionality to the core project` which is still precisely informative. Would you mind amending my commit and reword as desired - the commit author will be still preserved I believe.

Thanks!

Victor

weavejester:
Okay, done!

(It took me a while to figure out how to update the repository. Can't do it through SSH, has to be done through HTTPS, and if you have 2FA you can't use your password, but have to generate a developer token and use that instead.)

vemv:
> Can't do it through SSH, has to be done through HTTPS, and if you have 2FA you can't use your password, but have to generate a developer token and use that instead.)

Handy, thanks!

Looking forward to a new release 🍻

PR #128, opened 2018-07-22
Fixed bug where incorrect file path separator would be used on windows

xTVaser:
When running `lein cljfmt check` on Windows, regex exceptions occur.  For example:
```
java.util.regex.PatternSyntaxException: Unexpected internal error near index 1
\
 ^
 at java.util.regex.Pattern.error (:-1)
    java.util.regex.Pattern.compile (:-1)
    java.util.regex.Pattern.<init> (:-1)
    java.util.regex.Pattern.compile (:-1)
    clojure.core$re_pattern.invokeStatic (core.clj:4665)
    clojure.core$re_pattern.invoke (core.clj:4657)
    cljfmt.diff$to_absolute_path.invokeStatic (diff.clj:14)
```

This is caused by the use of `File/separator` in Windows environments.  
https://github.com/weavejester/cljfmt/blob/624ab9649818b11ad9806590a2ed0ba181c5b2a6/cljfmt/src/cljfmt/diff.clj#L13-L16

Despite the path being retrieved from a `URI` object
https://github.com/weavejester/cljfmt/blob/624ab9649818b11ad9806590a2ed0ba181c5b2a6/cljfmt/src/cljfmt/main.clj#L21-L24

On Windows, file paths separate on `\` but URLs _always_ use `/` - https://docs.oracle.com/javase/7/docs/api/java/net/URI.html?is-external=true

Therefore `/` can be used freely so that `lein cljfmt check` can execute successfully on Windows as well as Unix environments.

### DIFF
--- a/cljfmt/src/cljfmt/diff.clj
+++ b/cljfmt/src/cljfmt/diff.clj
@@ -1,6 +1,7 @@
 (ns cljfmt.diff
   (:import [difflib DiffUtils Delta$TYPE]
-           [java.io File])
+           [java.io File]
+           [java.util.regex Pattern])
   (:require [clojure.java.io :as io]
             [clojure.string :as str]))
 
@@ -11,7 +12,7 @@
   (str/join "\n" ss))
 
 (defn to-absolute-path [filename]
-  (->> (str/split filename (re-pattern File/separator))
+  (->> (str/split filename (re-pattern (Pattern/quote File/separator)))
        (apply io/file)
        .getCanonicalPath))
 
--- a/cljfmt/src/cljfmt/main.clj
+++ b/cljfmt/src/cljfmt/main.clj
@@ -19,9 +19,9 @@
     (apply println args)))
 
 (defn- relative-path [dir file]
-  (-> (.toURI dir)
-      (.relativize (.toURI file))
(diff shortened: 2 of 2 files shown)


xTVaser:
Agreed, I had just assumed that a `URI` was chosen for particular reason and went with the least-invasive change.  I can look into changing things so that `to-absolute-path` gets a proper filepath, rather than a URL.

weavejester:
Thanks for the update. I don't think Github notifies the repository owner when there has been updates to code, only when there's been a new comment, so when you've made some changes you're happy with, can you add a quick comment to say you're done? I assume the commit you have is fine to be reviewed? You've tested that it works correctly under Windows, now?

xTVaser:
Sorry about that, yes the PR is ready for review.  Re-tested and the problem is fixed on Windows, and cljfmt continues to work on *nix systems (I used macOS).

khellang:
Any update on this? We're hitting it as well 😞

weavejester:
Ah, looks like I merged another PR that purported to do the same thing, though it lacks some of the code this PR does. The commit message in this PR looks like it needs to be updated, so I'll add the extra code in manually.

PR #130, opened 2018-07-30
Indent clojure.spec.alpha/fdef like clojure.core/def

JamesLaverack:
Clojure Spec's `fdef` is currently indented by cljfmt like any other form:

```
(s/fdef update-balance
        :args (s/cat :entry :ccm.transaction/entry
                     :balance :ccm.account/balance)
        :ret :ccm.account/balance
        :fn #(= (-> % :ret :ccm.money/currency)
                (-> % :args second :ccm.money/currency)))
```

However, both the [Clojure spec guide](https://clojure.org/guides/spec#_spec_ing_functions) and [Emacs clojure-mode](https://github.com/clojure-emacs/clojure-mode/blob/master/CHANGELOG.md#580-2018-06-26) indent it like a `clojure.core/def`.

This pull request brings cljfmt in-line with the `def`-style formatting and indents like this:

```
(s/fdef ranged-rand
  :args (s/and (s/cat :start int? :end int?)
               #(< (:start %) (:end %)))
  :ret int?
  :fn (s/and #(>= (:ret %) (-> % :args :start))
             #(< (:ret %) (-> % :args :end))))
```

Note that as a workaround you can currently specify a rule for `fdef` in your `project.clj` (if using `lein-cljfmt`). This is what we do internally:

```
:cljfmt {:indents {fdef [[:inner 0]]}}
```

### DIFF
--- a/cljfmt/resources/cljfmt/indents/clojure.clj
+++ b/cljfmt/resources/cljfmt/indents/clojure.clj
@@ -30,6 +30,7 @@
  extend          [[:block 1]]
  extend-protocol [[:block 1] [:inner 1]]
  extend-type     [[:block 1] [:inner 1]]
+ fdef            [[:inner 0]]
  finally         [[:block 0]]
  fn              [[:inner 0]]
  for             [[:block 1]]
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -35,7 +35,17 @@
     (is (= (reformat-string "(fn [x]\n(foo bar\nbaz))")
            "(fn [x]\n  (foo bar\n       baz))"))
     (is (= (reformat-string "(fn [x] (foo bar\nbaz))")
-           "(fn [x] (foo bar\n             baz))")))
+           "(fn [x] (foo bar\n             baz))"))
+    (is (= (reformat-string (str "(clojure.spec.alpha/def ::foo\n"
+                                 "                        string?)" ))
+           "(clojure.spec.alpha/def ::foo\n  string?)"))
+    (is (= (reformat-string
+            (str "(clojure.spec.alpha/fdef foo\n"
+                 "                         :args (clojure.spec.alpha/cat :x string?)\n"
+                 "                         :ret nat-int?)"))
+           (str "(clojure.spec.alpha/fdef foo\n"
+                "  :args (clojure.spec.alpha/cat :x string?)\n"
+                "  :ret nat-int?)"))))
 
   (testing "inner indentation"
     (is (= (reformat-string "(letfn [(foo [x]\n(* x x))]\n(foo 5))")


weavejester:
This seems fine, I think. Can you remove the backticks from the commit message?

JamesLaverack:
Sure thing. I've force-pushed a commit with an updated message.

I've updated the PR name too so you don't get the backticks in the merge commit message.

weavejester:
Thanks! I've also just noticed that the test lines are a little long, as all the other lines in the test namespace use `str` to split up long sections. Can you split your test up in the same way? Sorry I didn't catch this before.

JamesLaverack:
Something more like this maybe?

```
    (is (= (reformat-string (str "(clojure.spec.alpha/def ::foo\n"
                                 "string?)" ))
           "(clojure.spec.alpha/def ::foo\n  string?)"))
    (is (= (reformat-string (str "(clojure.spec.alpha/fdef foo\n"
                                 ":args (clojure.spec.alpha/cat :x string?)\n"
                                 ":ret nat-int?)"))
           (str "(clojure.spec.alpha/fdef foo\n"
                "  :args (clojure.spec.alpha/cat :x string?)\n"
                "  :ret nat-int?)"))))
```

This does change the semantics of the test slightly in that it no longer has large amounts of whitespace. However the rest of the tests in the file generally don't have inputs with too much whitespace either. So this is probably closest to existing style.

If you're happy with that I'll update the branch. Is another commit fine or would you prefer I rewrite the single existing commit for a cleaner history after merge?

weavejester:
> This does change the semantics of the test slightly in that it no longer has large amounts of whitespace.

You can still keep the whitespace in the strings, can't you? I have no preference either way, but if you want to add more whitespace to the inputs you can.

> Is another commit fine or would you prefer I rewrite the single existing commit for a cleaner history after merge?

The latter, please.

I see a commit history as a way for a developer to understand the evolution of the source code; commits that just change formatting don't make the history clearer, so I prefer to have them edited out when possible (i.e. in PR branches that haven't been merged into master yet).

JamesLaverack:
>You can still keep the whitespace in the strings, can't you? I have no preference either way, but if you want to add more whitespace to the inputs you can.

JamesLaverack:
> You can still keep the whitespace in the strings, can't you? I have no preference either way, but if you want to add more whitespace to the inputs you can.

JamesLaverack:
> You can still keep the whitespace in the strings, can't you? I have no preference either way, but if you want to add more whitespace to the inputs you can.

That's true. It makes the lines a bit longer even with the `str`, but I agree it makes it more readable.

> I see a commit history as a way for a developer to understand the evolution of the source code; commits that just change formatting don't make the history clearer, so I prefer to have them edited out when possible (i.e. in PR branches that haven't been merged into master yet).

Makes sense. I've amended the commit with the updates.

PR #138, opened 2018-10-08
Change linebreak handling to work with Windows (fixes #137)

### LINKED ISSUE #137: line endings issue on Windows
cljfmt will (incorrectly) flag the line ending as a change to be made after running `lein cljfmt check` when run on Windows (line endings = `\r\n`). cljfmt's corrected version uses `\n` as line ending.

image

running `lein cljfmt fix` and then running `git diff`:

image

`\n` results in `LF` being written to file
`\r\n` results in `CR` `LF` being written to file

sloxy:
Linebreaks normalised to *nix '\n' (LF) for comparison.

Linebreak in re-written file will be same as linebreak in original file.
If no linebreak in original file, linebreak in rewritten will be OS default.

short vid showing it running on Win OS and handling both `LF` `(\n)` and `CR``LF` `(\n)` linebreaks: https://www.youtube.com/watch?v=OWpzFekl49U

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -2,6 +2,7 @@
   #?@(:clj
       [(:refer-clojure :exclude [reader-conditional?])
        (:require [clojure.java.io :as io]
+                 [clojure.string :as str]
                  [clojure.zip :as zip]
                  [rewrite-clj.node :as n]
                  [rewrite-clj.parser :as p]
@@ -390,3 +391,20 @@
          (reformat-form (cond-> options
                           alias-map (assoc :alias-map alias-map)))
          (n/string)))))
+
+(def default-line-separator
+  #?(:clj (System/lineSeparator) :cljs \newline))
+
+(defn normalize-newlines [s]
+  (str/replace s #"\r\n" "\n"))
+
+(defn replace-newlines [s sep]
+  (str/replace s #"\n" sep))
+
+(defn find-line-separator [s]
+  (or (re-find #"\r?\n" s) default-line-separator))
+
+(defn wrap-normalize-newlines [f]
+  (fn [s]
+    (let [sep (find-line-separator s)]
+      (-> s normalize-newlines f (replace-newlines sep)))))
--- a/cljfmt/src/cljfmt/main.clj
+++ b/cljfmt/src/cljfmt/main.clj
@@ -34,7 +34,7 @@
       [f])))
 
 (defn- reformat-string [options s]
-  (cljfmt/reformat-string s options))
+  ((cljfmt/wrap-normalize-newlines #(cljfmt/reformat-string % options)) s))
 
 (defn- project-path [{:keys [project-root]} file]
   (-> project-root (or ".") io/file (relative-path (io/file file))))
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -1,7 +1,9 @@
 (ns cljfmt.core-test
   (:require [#?@(:clj (clojure.test :refer)
                  :cljs (cljs.test :refer-macros)) [deftest testing is]]
-            [cljfmt.core :refer [reformat-string]]
+            [cljfmt.core :refer [reformat-string default-line-separator
+                                 normalize-newlines find-line-separator
+                                 replace-newlines wrap-normalize-newlines]]
             [clojure.string :as str]))
 
 (deftest test-indent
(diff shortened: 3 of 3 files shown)


cljfmt/src/cljfmt/core.cljc
```diff
@@ -390,3 +391,19 @@
...
                           alias-map (assoc :alias-map alias-map)))
          (n/string)))))
+
+(def default-linebreak
+  #?(:clj (System/lineSeparator) :cljs \newline))
+
+(defn linebreak-type [text]
+  "Get the newline character from text.
+   If text doesn't have newline, use default."
+  (if-let [newline (re-find #"\r?\n" text)]
```
weavejester: This could be written more concisely as:

```clojure
(or (re-find #"\r?\n" text) default-linebreak)
```

cljfmt/src/cljfmt/core.cljc
```diff
@@ -390,3 +391,19 @@
          (reformat-form (cond-> options
                           alias-map (assoc :alias-map alias-map)))
          (n/string)))))
+
+(def default-linebreak
+  #?(:clj (System/lineSeparator) :cljs \newline))
+
+(defn linebreak-type [text]
```
weavejester: The docstring can be ommitted since this is an internal function, and it should be moved to `cljfmt.diff`. Also call it something like `find-line-separator`.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -390,3 +391,19 @@
          (reformat-form (cond-> options
                           alias-map (assoc :alias-map alias-map)))
          (n/string)))))
+
+(def default-linebreak
```
weavejester: Maybe rename this to `default-line-separator`, and move it to `cljfmt.diff` where it's being used.

cljfmt/src/cljfmt/diff.clj
```diff
@@ -2,14 +2,15 @@
...
+            [clojure.java.io :as io]
             [clojure.string :as str]))
 
 (defn- lines [s]
-  (str/split s #"\n"))
+  (str/split-lines s))
 
-(defn- unlines [ss]
-  (str/join "\n" ss))
+(defn- unlines [ss lb]
```
weavejester: `unlines` can be omitted if it's just an alias for `str/join`.

cljfmt/src/cljfmt/diff.clj
```diff
@@ -2,14 +2,15 @@
...
            [java.io File]
            [java.util.regex Pattern])
-  (:require [clojure.java.io :as io]
+  (:require [cljfmt.core :as cljfmt]
+            [clojure.java.io :as io]
             [clojure.string :as str]))
 
 (defn- lines [s]
-  (str/split s #"\n"))
+  (str/split-lines s))
```
weavejester: `lines` can be omitted if it's just an alias for `str/split-lines`.

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -283,3 +283,20 @@
...
+  (is (= (linebreak-type "foo\\nbar\\r\nbaz") "\n"))
+  (is (= (linebreak-type "foobarbaz") cljfmt.core/default-linebreak)))
+
+(deftest test-linebreak-replacement
+  (is (= (linebreak-replaced "foo\nbar\nbaz" "\n") "foo\nbar\nbaz"))
+  (is (= (linebreak-replaced "foo\nbar\nbaz" "\r\n") "foo\r\nbar\r\nbaz"))
+  (is (= (linebreak-replaced "foo\r\nbar\r\nbaz" "\n") "foo\nbar\nbaz"))
+  (is (= (linebreak-replaced "foo\r\nbar\r\nbaz" "\r\n") "foo\r\nbar\r\nbaz"))
+  (is (= (linebreak-replaced "foobarbaz" "\n") "foobarbaz"))
+  (is (= (linebreak-replaced "foobarbaz" "\r\n") "foobarbaz")))
```
weavejester: Missing newline at end of document.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -390,3 +391,19 @@
...
+   If text doesn't have newline, use default."
+  (if-let [newline (re-find #"\r?\n" text)]
+    newline
+    default-linebreak))
+
+(defn linebreak-replaced [text linebreak]
+  "Substitues in 'linebreak' as newline character(s)"
+  (if (= linebreak (linebreak-type text))
+    text
+    (->> text (str/split-lines) (str/join linebreak))))
```
weavejester: Missing newline at end of document.

cljfmt/src/cljfmt/core.cljc
sloxy: renamed. it's also being used indirectly by main via (now renamed) `find-line-separator`

cljfmt/src/cljfmt/core.cljc
sloxy: renamed. it's used by both `diff` & `main` so will leave it in core.

sloxy:
i have moved `linebreak-replaced` from `core` into `main` as it is only used there.
am i alright to move the corresponding tests to a new `main_test.clj` file and leave the function as `defn`
or should i skip adding tests for `main.clj` and set the function as `defn-`

weavejester:
You should move the functions into the namespaces that need them. As far as I can see, `cljfmt.core` doesn't need to be touched by this PR.

sloxy:
diff & main tests - do you want these removed?

weavejester:
Could you tell me a little more about the problem that this PR will solve? Does the problem occur only when the file being formatted has a different line separator to the OS default?

sloxy:
> Could you tell me a little more about the problem that this PR will solve?

The issue is running `cljfmt` on a windows machine will flag the windows line endings(CR LF) as something that isn't formatted correctly.
(images in https://github.com/weavejester/cljfmt/issues/137)

> Does the problem occur only when the file being formatted has a different line separator to the OS default?

No, this problem occurs if the file does not have LF as linebreak character (regardless of OS). 
As Windows uses CR LF as linebreak chars, this issue manifests on Windows.

weavejester:
Okay, so perhaps we should look at a more general solution to this. Let's put the following in `core`, since it's a broader use-case than I previously understood.

```clojure
(defn normalize-newlines [s]
  (str/replace s #"\r\n" "\n"))

(defn replace-newlines [s sep]
  (str/replace s #"\n" sep))

(defn find-line-separator [s]
  (or (re-find #"\r?\n" s) (System/lineSeparator))

(defn wrap-normalize-newlines [f]
  (fn [s]
    (let [sep (find-line-separator s)]
      (-> s normalize-newlines f (replace-newlines sep)))))
```

Then in `main` we can use:

```clojure
(defn reformat-string [s options]
  ((wrap-normalize-newlines #(cljfmt/reformat-string % options)) s))
```

I don't think we need to touch `diff` if we're updating `main`, do we?

sloxy:
updated. let me know if further changes req'd.

cljfmt/src/cljfmt/main.clj
```diff
@@ -34,8 +34,7 @@
       [f])))
 
 (defn- reformat-string [options s]
-  (cljfmt/reformat-string s options))
-
+  ((cljfmt/wrap-normalize-newlines #(cljfmt/reformat-string % options)) s))
```
weavejester: Can you put the missing newline back in?

weavejester:
Thanks, that looks almost perfect. Can you put back the newline that was accidentally deleted in the commit, and slightly change the commit message to:

```
Change linebreak handling to work with Windows

Linebreaks normalised to *nix '\n' (LF) for comparison.

Linebreak in re-written file will be same as linebreak in original file.
If no linebreak in original file, linebreak in rewritten will be OS
default.

Fixes #137.
```

This fixes the line widths (50 characters for subject, 72 characters for body).

khellang:
Thanks for jumping on these @weavejester 😄

cljfmt/src/cljfmt/main.clj
sloxy: 👍 sorry didn't clock it last night. updating now

PR #143, opened 2018-11-05
Check for not command-next at indent to fix remove-surrounding-whites…

### LINKED ISSUE #141: Some lines with only a closing bracket get no indent at all, when not removing surrounding whitespace
I'm using `cljfmt/reformat-string` with this option map:

```clojure
{:remove-surrounding-whitespace? false}
```

Formatting:
```clojure
(foo
(bar
)
)
```

I expect:
```clojure
(foo
 (bar
  )
 )
```

But I get:
```clojure
(foo
 (bar
)
 )
```

Am I misunderstanding something or is my expectation correct?

Formatting:
```clojure
(foo
(bar
))
```

Results in the expected:
```clojure
(foo
 (bar
  ))
```

EDIT: I'm using cljfmt from ClojureScript, if that has anything to do with anything.

pfeodrippe:
…pace? set to false

Tries to fix #141.

I'm still puzzled about the function `line-break-next?` though, @weavejester, is there some test I could add that invalidates this PR?

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -123,11 +123,8 @@
 (defn- comment-next? [zloc]
   (-> zloc zip/next skip-whitespace comment?))
 
-(defn- line-break-next? [zloc]
-  (-> zloc zip/next skip-whitespace line-break?))
-
 (defn- should-indent? [zloc]
-  (and (line-break? zloc) (not (line-break-next? zloc))))
+  (and (line-break? zloc) (not (comment-next? zloc))))
 
 (defn- should-unindent? [zloc]
   (and (indentation? zloc) (not (comment-next? zloc))))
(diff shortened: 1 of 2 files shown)


weavejester:
Thanks for the PR! I'm not sure what `line-break-next?` is used for. I thought it might be for ensuring comments are indented as well, but when I checked just now I discovered they're not indented in the master branch. If it passes all the tests, and I can't think of a failing test, then we'll assume it's okay.

However, for consistency can you swap the actual/expected order in the test you added so it's the same as the others. If we're going to reverse the order to the more common form I'd prefer that done all in one go. Can you also change the commit message so it adheres to the [seven rules](https://chris.beams.io/posts/git-commit/#seven-rules).

pfeodrippe:
@weavejester Thanks for the link! If it's not good yet, say so o/

weavejester:
Sorry for the delay in getting back to you. The commit message needs a little tweaking, as you've mispelled "indent" and "indentation" and the line lengths are incorrect. Here's a fixed version:

```
Fix indent when not remove surrounding whitespace

When :remove-surrounding-whistepace? is set to false, there were some
cases where the indentation was incorrect. This fixes the issue.

Resolves: #141.
```

If you amend your commit with that message I'll merge it in and release a new version.

pfeodrippe:
Don't worry, @weavejester, and thanks for the review =)

Keep the awesome work o/

PEZ:
Thanks for fixing two of ”my” issues, @pfeodrippe! ❤️

Issue #112, opened 2018-01-23
comments inside map literals aren't handled consistently


gonewest818:
Here's an example.  If you have code like this:

```
(def h{
;; comment
:this :is
    ;; comment
:a :multi
  ; comment
:line :hash-map
            ;;; comment
})
```

then `cljfmt fix` produces this:

```
(def h {;; comment
        :this :is
    ;; comment
        :a :multi
  ; comment
        :line :hash-map
            ;;; comment
})
```

where the hash-map entries are all aligned properly to the opening `{`. The interior comments seem to be ignored, however.  And incidentally, the alignment of the closing `})` seems identical to #111, but I'm not sure if what's happening with the comments is also related.

lread:
If I'm understanding things correctrly, the first comment moves due to `:remove-surrounding-whitespace?` being `true`. The rest of the comments remain untouched. I'm wondering what the expected behavior is here.  

I agree that the closing `})` is a dupe of #111 and probably also #82.

lread:
If we look at conventions, we use [two semicolons for code fragments](https://github.com/bbatsov/clojure-style-guide#two-semicolons-for-code-fragment).  So the double semicolon comment lines should align with the fragment.  

[Triple semicolons are are for top level items](https://github.com/bbatsov/clojure-style-guide#three-semicolons-for-top-level-comments)  and really should not be in code, so they should remain untouched.  

The [single semicolon is for margin comments](https://github.com/bbatsov/clojure-style-guide#one-semicolon-for-margin-comments), for now I propose we leave them untouched.  

So I think that means the original example given:
```clojure
(def h{
;; comment
:this :is
    ;; comment
:a :multi
  ; comment
:line :hash-map
            ;;; comment
})
```
should format like so:
```clojure
(def h {;; comment
        :this :is
        ;; comment
        :a :multi
  ; comment
        :line :hash-map
            ;;; comment
        })
```

lread:
@weavejester, when you have a moment, what do you think?

lread:
Thanks @weavejester, I think I'll wait to see if there is interest from others before working on this then.  Given [this guidance](https://github.com/bbatsov/clojure-style-guide#two-semicolons-for-code-fragment) I thought the double semicolon might be a good candidate for indenting by default, but maybe its usage as a code fragment comment is not as idiomatic as we'd like?  

[...]

Issue #111, opened 2018-01-22
Parens after trailing comment not indented


arrdem:
I encountered this bug while trying to format a codebase. It looks like the trailing comment causes the computation of the last indentation width not to occur correctly. If there's a trailing newline, it seems that the computed indentation width is `0` somehow, otherwise the left margin of the previous form seems to win.

Neither of these is expected behavior, I'd expect that the closing stack of parenthesis always be indented to the body width of the let being closed.

## Ex 1. Close paren stacks after comments not correctly indented

```clojure
cljfmt.core> (print (reformat-string "(deftest foo
  (testing \"testing foo\"
    (let []
      ;; this messes up
      )))"))
(deftest foo
  (testing "testing foo"
    (let []
      ;; this messes up
         )))
nil
```

## Ex 2. Newline after paren stack impacts behavior

```
cljfmt.core> (print (reformat-string "(deftest foo
  (testing \"testing foo\"
    (let []
      ;; this messes up
      )))
"))
(deftest foo
  (testing "testing foo"
    (let []
      ;; this messes up
)))
nil
cljfmt.core> 
```

MatthewDarling:
Seems related to #82. Which you had said in ~16 months ago "shouldn't be too hard to fix", haha.

At least, Ex 2 seems the same as that issue. Ex 1 is new to me, but it seems likely to be the same underlying problem.

When I was working on it, it was proving difficult because parens don't actually have nodes in the tree. Although, thinking about it now - I wonder if end-of-line comments could have indentation logic which inserts whitespace to the tree after themselves? Depending on how `#_` and `(comment ... )` are parsed, might have to treat them differently from semi-colon comments.

arrdem:
Argh 😄 well, I guess it's time for me to eat my words. I'll see if I can pick this up today.

dijonkitchen:
Should this be closed since it's a duplicate of #82?

lread:
This is a slight variation on #82 in that the comment line is the only line in the indent block.

lread:
At the time of this writing example 1 still behaves as reported but example 2 now behaves the same as example 1.  This currently seems to be an issue with indent blocks with no code in them.

PR #146, opened 2018-11-19
Make string-format unit tests more readable

### LINKED ISSUE #145: Make unit tests easier to understand
I was about to dig into an open issue when I found that I was having difficulty reading reformat-string unit tests, hence this proposal. 

Inspired by the conventions in #88, I think we can make reformat-string unit tests easier to digest by:
1. Giving each line its own line (no longer embedding \n)
1. Making the intent of each test clear by giving it a description. Having tests separated into testing block helped, but sometimes I wondered what an individual test was trying to achieve.
1. Making failures easier to decipher. When a reformat-string test failed, I was left confused between actual and expected results; they seem reversed. Note that I am using [eftest](https://github.com/weavejester/eftest) for friendlier test output.

I'm using this issue to propose the change. It seems like a good idea to me, what do you think? If you like it, I'm happy to do the work.

# Simple test example
## Code
Current
```clojure
    (is (= (reformat-string "(foo bar\nbaz\nquz)")
           "(foo bar\n     baz\n     quz)"))
```
Proposed new 
```clojure
    (rft "hangs indent to deeper element"
        ["(foo bar"
         "baz"
         "quz)"]
        ["(foo bar"
         "     baz"
         "     quz)"])
```
At a glance, there is no doubt about intended indentation.

## Failures
Suppose I changed `baz` to `biz` in above example for the expected result. Using eftest here's what failures look like.

Current
```
FAIL in cljfmt.core-test/test-indent (core_test.cljc:43)
list indentation
expected: "(foo bar\n     baz\n     quz)"
  actual: ("(foo bar\n     biz\n     quz)")
```
Expected and actual seem reversed to me.

Proposed new 
```
FAIL in cljfmt.core-test/test-indent (core_test.cljc:16)
list indentation
hangs indent to deeper element
expected: ["(foo bar" "     biz" "     quz)"]
  actual: (["(foo bar" "     baz" "     quz)"])
```
Expected and actual are in correct order and test description is shown.

# Comments tests
To get a feel for what these would like, let's dive into a testing block.
Here's what comments tests currently look like:
```clojure
  (testing "comments"
    (is (= (reformat-string ";foo\n(def x 1)")
           ";foo\n(def x 1)"))
    (is (= (reformat-string "(ns foo.core)\n\n;; foo\n(defn foo [x]\n(inc x))")
           "(ns foo.core)\n\n;; foo\n(defn foo [x]\n  (inc x))"))
    (is (= (reformat-string ";; foo\n(ns foo\n(:require bar))")
           ";; foo\n(ns foo\n  (:require bar))"))
    (is (= (reformat-string "(defn foo [x]\n  ;; +1\n(inc x))")
           "(defn foo [x]\n  ;; +1\n  (inc x))"))
    (is (= (reformat-string "(let [;foo\n x (foo bar\n baz)]\n x)")
           "(let [;foo\n      x (foo bar\n             baz)]\n  x)"))
    (is (= (reformat-string "(binding [x 1] ; foo\nx)")
           "(binding [x 1] ; foo\n  x)")))
```
Here's my proposal for change:
```clojure
  (testing "comments"
    (rft "header comment unchanged"
        [ ";foo"
          "(def x 1)"]
        [ ";foo"
         "(def x 1)"])
    (rft "post namespace comment unchanged, code reformatted"
        ["(ns foo.core)"
         ""
         ";; foo"
         "(defn foo [x]"
         "(inc x))"]
        ["(ns foo.core)"
         ""
         ";; foo"
         "(defn foo [x]"
         "  (inc x))"])
    (rft "pre namespace comment unchanged, namespace reformatted"
        [";; foo"
         "(ns foo"
         "(:require bar))"]
        [";; foo"
         "(ns foo"
         "  (:require bar))"])
    (rft "commented out line unchanged, code reformatted"
        ["(defn foo [x]"
         "  ;; +1"
         "(inc x))"]
        ["(defn foo [x]"
         "  ;; +1"
         "  (inc x))"])
    (rft "inline comment unchanged, code reformatted"
        ["(let [;foo"
         " x (foo bar"
         " baz)]"
         " x)"]
        ["(let [;foo"
         "      x (foo bar"
         "             baz)]"
         "  x)"])
    (rft "binding inline comment unchanged, code reformated"
        ["(binding [x 1] ; foo"
         "x)"]
        ["(binding [x 1] ; foo"
         "  x)"]))
```
# Example with options
Here's how I propose a reformat-string test with options would be expressed:
```clojure
    (rft "custom indents applied to namespaced defrecord"
         ["(ns example"
          "(:require [thing.core :as t]))"
          ""
          "(t/defrecord Foo [x]"
          "Closeable"
          "(close [_]"
          "(prn x)))"
          ""
          "(defrecord Foo [x]"
          "Closeable"
          "(close [_]"
          "(prn x)))"]
         {:indents {'thing.core/defrecord [[:inner 0]]}
          #?@(:cljs [:alias-map {"t" "thing.core"}])}
         ["(ns example"
          "    (:require [thing.core :as t]))"
          ""
          "(t/defrecord Foo [x]"
          "  Closeable"
          "  (close [_]"
          "         (prn x)))"
          ""
          "(defrecord Foo [x]"
          "           Closeable"
          "           (close [_]"
          "                  (prn x)))"])
```

lread:
Rework string-format unit tests to be easier to digest by:

 - Giving each input line its own line (no more embbeded \n)

 - Making the intent of each test clear by giving it a description

 - Making failures easier to decipher by switching expected and actual

Fixes #145.

### DIFF
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -4,289 +4,738 @@
             [cljfmt.core :refer [reformat-string default-line-separator
                                  normalize-newlines find-line-separator
                                  replace-newlines wrap-normalize-newlines]]
-            [clojure.string :as str]))
+            [cljfmt.test-util.common :as common]
+            #?(:clj [cljfmt.test-util.clojure]))
+  #?(:cljs (:require-macros [cljfmt.test-util.cljs])))
 
 (deftest test-indent
   (testing "list indentation"
-    (is (= (reformat-string "(foo bar\nbaz\nquz)")
-           "(foo bar\n     baz\n     quz)"))
-    (is (= (reformat-string "(foo\nbar\nbaz)")
-           "(foo\n bar\n baz)")))
+    (is (reformats-to?
+         ["(foo bar"
+          "baz"
+          "quz)"]
+         ["(foo bar"
+          "     baz"
+          "     quz)"]))
+
+    (is (reformats-to?
+         ["(foo"
+          "bar"
+          "baz)"]
+         ["(foo"
+          " bar"
+          " baz)"])
+        "to first arg"))
 
   (testing "block indentation"
-    (is (= (reformat-string "(if (= x 1)\n:foo\n:bar)")
-           "(if (= x 1)\n  :foo\n  :bar)"))
-    (is (= (reformat-string "(do\n(foo)\n(bar))")
-           "(do\n  (foo)\n  (bar))"))
-    (is (= (reformat-string "(do (foo)\n(bar))")
-           "(do (foo)\n    (bar))"))
-    (is (= (reformat-string "(deftype Foo\n[x]\nBar)")
-           "(deftype Foo\n         [x]\n  Bar)"))
-    (is (= (reformat-string "(cond->> x\na? a\nb? b)")
-           "(cond->> x\n  a? a\n  b? b)")))
+    (is (reformats-to?
+         ["(if (= x 1)"
+          ":foo"
+          ":bar)"]
+         ["(if (= x 1)"
+          "  :foo"
+          "  :bar)"]))
+    (is (reformats-to?
+         ["(do"
+          "(foo)"
+          "(bar))"]
+         ["(do"
+          "  (foo)"
+          "  (bar))"]))
+    (is (reformats-to?
+         ["(do (foo)"
+          "(bar))"]
+         ["(do (foo)"
+          "    (bar))"])
+        "do block to hanging argument")
+    (is (reformats-to?
+         ["(deftype Foo"
+          "[x]"
+          "Bar)"]
+         ["(deftype Foo"
+          "         [x]"
+          "  Bar)"])
+        "deftype fields to class name")
+    (is (reformats-to?
+         ["(cond->> x"
+          "a? a"
+          "b? b)"]
+         ["(cond->> x"
+          "  a? a"
+          "  b? b)"])))
 
   (testing "constant indentation"
-    (is (= (reformat-string "(def foo\n\"Hello World\")")
-           "(def foo\n  \"Hello World\")"))
-    (is (= (reformat-string "(defn foo [x]\n(+ x 1))")
-           "(defn foo [x]\n  (+ x 1))"))
-    (is (= (reformat-string "(defn foo\n[x]\n(+ x 1))")
-           "(defn foo\n  [x]\n  (+ x 1))"))
-    (is (= (reformat-string "(defn foo\n([] 0)\n([x]\n(+ x 1)))")
-           "(defn foo\n  ([] 0)\n  ([x]\n   (+ x 1)))"))
-    (is (= (reformat-string "(fn [x]\n(foo bar\nbaz))")
-           "(fn [x]\n  (foo bar\n       baz))"))
-    (is (= (reformat-string "(fn [x] (foo bar\nbaz))")
-           "(fn [x] (foo bar\n             baz))"))
-    (is (= (reformat-string (str "(clojure.spec.alpha/def ::foo\n"
-                                 "                        string?)"))
-           "(clojure.spec.alpha/def ::foo\n  string?)"))
-    (is (= (reformat-string
-            (str "(clojure.spec.alpha/fdef foo\n"
-                 "                         :args (clojure.spec.alpha/cat :x string?)\n"
-                 "                         :ret nat-int?)"))
-           (str "(clojure.spec.alpha/fdef foo\n"
-                "  :args (clojure.spec.alpha/cat :x string?)\n"
-                "  :ret nat-int?)"))))
+    (is (reformats-to?
+         ["(def foo"
+          "\"Hello World\")"]
+         ["(def foo"
+          "  \"Hello World\")"]))
+    (is (reformats-to?
+         ["(defn foo [x]"
+          "(+ x 1))"]
+         ["(defn foo [x]"
+          "  (+ x 1))"])
+        "defn does not indent to hanging name and params")
+    (is (reformats-to?
+         ["(defn foo"
+          "[x]"
+          "(+ x 1))"]
+         ["(defn foo"
+          "  [x]"
+          "  (+ x 1))"])
+        "defn does not indent to hanging name")
+    (is (reformats-to?
+         ["(defn foo"
+          "([] 0)"
+          "([x]"
+          "(+ x 1)))"]
+         ["(defn foo"
+          "  ([] 0)"
+          "  ([x]"
+          "   (+ x 1)))"])
+        "defn multi-arity")
+    (is (reformats-to?
+         ["(fn [x]"
+          "(foo bar"
+          "baz))"]
+         ["(fn [x]"
+          "  (foo bar"
+          "       baz))"])
+        "fn with body")
+    (is (reformats-to?
+         ["(fn [x] (foo bar"
+          "baz))"]
+         ["(fn [x] (foo bar"
+          "             baz))"])
+        "fn with hanging body")
+    (is (reformats-to?
+         ["(clojure.spec.alpha/def ::foo"
+          "                        string?)"]
+         ["(clojure.spec.alpha/def ::foo"
+          "  string?)"])
+        "unhangs hanging spec indentation")
+    (is (reformats-to?
+         ["(clojure.spec.alpha/fdef foo"
+          "                         :args (clojure.spec.alpha/cat :x string?)"
+          "                         :ret nat-int?)"]
+         ["(clojure.spec.alpha/fdef foo"
+          "  :args (clojure.spec.alpha/cat :x string?)"
+          "  :ret nat-int?)"])
+        "unhangs hanging spec indentations"))
 
   (testing "inner indentation"
-    (is (= (reformat-string "(letfn [(foo [x]\n(* x x))]\n(foo 5))")
-           "(letfn [(foo [x]\n          (* x x))]\n  (foo 5))"))
-    (is (= (reformat-string "(reify Closeable\n(close [_]\n(prn :closed)))")
-           "(reify Closeable\n  (close [_]\n    (prn :closed)))"))
-    (is (= (reformat-string "(defrecord Foo [x]\nCloseable\n(close [_]\n(prn x)))")
-           "(defrecord Foo [x]\n  Closeable\n  (close [_]\n    (prn x)))")))
+    (is (reformats-to?
+         ["(letfn [(foo [x]"
+          "(* x x))]"
+          "(foo 5))"]
+         ["(letfn [(foo [x]"
+          "          (* x x))]"
+          "  (foo 5))"]))
+    (is (reformats-to?
+         ["(reify Closeable"
+          "(close [_]"
+          "(prn :closed)))"]
+         ["(reify Closeable"
+          "  (close [_]"
+          "    (prn :closed)))"]))
+    (is (reformats-to?
+         ["(defrecord Foo [x]"
+          "Closeable"
+          "(close [_]"
+          "(prn x)))"]
+         ["(defrecord Foo [x]"
+          "  Closeable"
+          "  (close [_]"
+          "    (prn x)))"])))
 
   (testing "data structure indentation"
-    (is (= (reformat-string "[:foo\n:bar\n:baz]")
-           "[:foo\n :bar\n :baz]"))
-    (is (= (reformat-string "{:foo 1\n:bar 2}")
-           "{:foo 1\n :bar 2}"))
-    (is (= (reformat-string "#{:foo\n:bar\n:baz}")
-           "#{:foo\n  :bar\n  :baz}"))
-    (is (= (reformat-string "{:foo [:bar\n:baz]}")
-           "{:foo [:bar\n       :baz]}")))
+    (is (reformats-to?
+         ["[:foo"
+          ":bar"
+          ":baz]"]
+         ["[:foo"
+          " :bar"
+          " :baz]"]))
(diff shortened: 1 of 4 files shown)


weavejester:
Thanks for the PR. I seem to recall putting `is` into a different function messes up the error stacktrace or line numbers, so I'd prefer we instead use `assert-expr`. Otherwise everything seems like a good idea. So perhaps instead of:

```clojure
(rft "map"
      ["{:foo 1"
       ":bar 2}"]
      ["{:foo 1"
       " :bar 2}"])
```

We write:

```clojure
(is (reformats-into?
     ["{:foo 1"
      ":bar 2}"]
     ["{:foo 1"
      " :bar 2}"]))
```

Via some `assert-expr` method like:

```clojure
(defmethod assert-expr 'reformats-into? [msg [input-lines expected-lines options]]
  `(let [input#    (str/join "\n" ~input-lines)
         actual#   (reformat-string input# ~(or options {}))
         expected# (str/join "\n" ~expected-lines)
         result#   (= expected# actual#)]
     (if result#
       (do-report {:type :pass, :message ~msg,
                   :expected expected#, :actual actual#})
       (do-report {:type :fail, :message ~msg,
                   :expected expected#, :actual actual#}))
     result#)))
```

lread:
Thanks for your thoughtful and kind review!  In addition to dealing with error stack trace line number issues, I think your test syntax is is more expressive than my original shot at it. I will take a look at making this change soon.

lread:
My macro fu leveled up slightly. [Portable Custom clojure.test Assertions](http://seespotcode.net/2018/01/13/portable-clojure-test-assert-expr/) helped me quite a bit!

For your consideration:
1. I kept my comparison on vectors of lines because I preferred its output on error over a single string (I assume you might have been suggesting single string from your `assert-expr` sample in above comment.) Here's what an error looks like using eftest: 
```
FAIL in cljfmt.core-test/test-indent (core_test.cljc:233)
namespaces
applies custom indentation to namespaced defrecord
expected: ["(ns example"
           "    (:require [thing.core :as t]))"
           ""
           "(t/defrecord Boo [x]"
           "  Closeable"
           "  (close [_]"
           "         (prn x)))"
           ""
           "(defrecord Foo [x]"
           "           Closeable"
           "           (close [_]"
           "                  (prn x)))"]
  actual: ["(ns example"
           "    (:require [thing.core :as t]))"
           ""
           "(t/defrecord Foo [x]"
           "  Closeable"
           "  (close [_]"
           "         (prn x)))"
           ""
           "(defrecord Foo [x]"
           "           Closeable"
           "           (close [_]"
           "                  (prn x)))"]
```
2. I chose the shorter `reformats-to?` over suggested `reformats-into?`.
3. My original commit forced a description on each test. While reworking to assert expressions, I realized that this was overkill and only provided descriptions where I felt they would genuinely help.

lread:
After we are all good and this is merged I was thinking of issues #82, #111 and #112.  I have taken a peek and it seems to me they have been almost (but not entirely) fixed by other changes. I'll add proper unit test coverage.

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -4,289 +4,744 @@
...
+      "discard")
+  (is (reformats-to?
+       ["(juxt +' -')"]
+       ["(juxt +' -')"])
+      "arbirtrary sum of sums and negation")
+  (is (reformats-to?
+       ["#\"(?i)foo\""]
+       ["#\"(?i)foo\""])
+      "regular expression")
+  ; newline is embedded in regex, this test is more readable if we don't use our assert wrapper
```
weavejester: This comment can be removed, as I don't think it's necessary to understand the code.

cljfmt/test/cljfmt/assert/cljs.cljc
```diff
@@ -0,0 +1,10 @@
+(ns cljfmt.assert.cljs
+  (:require [cljs.test :as test]
+            [cljfmt.assert.impl :as impl]))
+
+;; cljs macros are compiled by clj
+#?(:clj
```
weavejester: Why do you have `cljfmt.assert.cljs`, `cljfmt.assert.impl` and `cljfmt.asserts`? Can you put everything into `cljfmt.core-test`?

lread: My code was heavily inspired by [this example](https://github.com/grzm/ex.assertions) which splits things out in a very similar way. I personally prefer to move self contained complexity out to separate files.  I find when source files get larger they also get harder to digest and maintain. I realize this is a subjective call and am happy to move to your preference.

`asserts.cljc` could definitely be merged into `core_test.cljc`, but I'm thinking there might be some file separation gymnastics needed to support test assertions under ClojureScript: `assert/cljs.cljc` is required for ClojureScript but is compiled by Clojure. I am new to this area though, I'll try to understand what is possible and get back to you.

You did not ask but: I added the `assert` sub-directory to try to convey these are internal details that `core_test` need not concern itself with directly.

weavejester: I'd rather just move the method into `core_test.cljc`, or possibly `test_util.cljc`. The way the cljs macros interact might make placing it in a secondary namespace a good idea. However, the namespace should indicate clearly that it's a test namespace, hence `test-util`. You should also be able to do it with only one `.cljc` file.

lread: I have not succeeding at merging everything into `core_test.cljc`. I did manage to merge `asserts.cljc` into `core_test.cljc`.  I'm left with the cljs `assert-expr` in a separate source file, and the code that the clj and cljs `assert-expr`s share in another file.  My last push called these `assert_cljs.cljc` and `assert_impl.cljc` respectively.  Do you think moving these two files under a `test-util` subdir would work for you?

BTW, thanks for your patience and encouragement, it is much appreciated.

lread: Thanks for clearing that up, I was thinking I was missing some concept or other.  

If I understand correctly, which I might not, `common` and `cljs` files need the `cljc` extension. 

Your new separation makes usage from ns in `core-tests` very clear.  I also like your tweaks to the code, you have a real talent for succinctness.

Note: I've left the order of args when including options as: `in-line options expected-lines`. I feel they belong next to `in-lines` for better readability of tests.

weavejester: > If I understand correctly, which I might not, `common` and `cljs` files need the `cljc` extension.

What's your reasoning? If `assert-expr` is used in `is`, and `is` is a macro, doesn't that mean that only Clojure files are necessary?

lread: I misled you, I don't understand yet. Let me rephrase: I do not know why, but to get the build to succeed, I needed to rename `common` and `cljs` files with the `cljc` extension.

weavejester: Okay, lets set them as `cljc` then.

lread: Ok, I've pushed changes, they are ready for your review.

cljfmt/test/cljfmt/test_util/common.cljc
```diff
@@ -0,0 +1,16 @@
...
+(defn assert-reformats-to
+  ([msg in-lines expected-lines]
+   (assert-reformats-to msg in-lines {} expected-lines))
+  ([msg in-lines options expected-lines]
+   (let [input (str/join "\n" in-lines)
+         actual (reformat-string input options)
+         preserve-trailing-empty-strings -1
+         actual-lines (str/split actual #"\n" preserve-trailing-empty-strings)]
+     {:type (if (= expected-lines actual-lines) :pass :fail)
+      :message msg,
```
weavejester: Can you remove this extra `,`

cljfmt/test/cljfmt/test_util/common.cljc
```diff
@@ -0,0 +1,16 @@
+(ns cljfmt.test-util.common
+  (:require [clojure.string :as str]
+            [cljfmt.core :refer [reformat-string]]))
+
+(defn assert-reformats-to
+  ([msg in-lines expected-lines]
+   (assert-reformats-to msg in-lines {} expected-lines))
+  ([msg in-lines options expected-lines]
```
weavejester: I think I'd prefer to have the options last. I understand why you put the `expected-lines` there, but I think the argument list is more predictable if the positioning doesn't change.

cljfmt/test/cljfmt/test_util/common.cljc
```diff
@@ -0,0 +1,16 @@
...
+   (let [input (str/join "\n" in-lines)
+         actual (reformat-string input options)
+         preserve-trailing-empty-strings -1
```
weavejester: I don't think we necessarily need this, since someone can just look up what `-1` means in this context.

cljfmt/test/cljfmt/test_util/common.cljc
```diff
@@ -0,0 +1,16 @@
+(ns cljfmt.test-util.common
+  (:require [clojure.string :as str]
+            [cljfmt.core :refer [reformat-string]]))
+
+(defn assert-reformats-to
+  ([msg in-lines expected-lines]
+   (assert-reformats-to msg in-lines {} expected-lines))
+  ([msg in-lines options expected-lines]
+   (let [input (str/join "\n" in-lines)
```
weavejester: Can you ensure the `let` fields are aligned?

cljfmt/test/cljfmt/test_util/common.cljc
```diff
@@ -0,0 +1,16 @@
...
+         actual (reformat-string input options)
+         preserve-trailing-empty-strings -1
+         actual-lines (str/split actual #"\n" preserve-trailing-empty-strings)]
+     {:type (if (= expected-lines actual-lines) :pass :fail)
```
weavejester: Can you ensure the map fields are aligned?

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -4,289 +4,750 @@
             [cljfmt.core :refer [reformat-string default-line-separator
                                  normalize-newlines find-line-separator
                                  replace-newlines wrap-normalize-newlines]]
-            [clojure.string :as str]))
+            [cljfmt.test-util.common :as common])
+  #?(:clj (:require [cljfmt.test-util.clojure])
+     :cljs (:require-macros [cljfmt.test-util.cljs])))
+
+;;; `reformats-to?` makes `reformat-string` tests more readable.
```
weavejester: I think these lines can be removed, as the tests themselves are good examples of use.

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -4,289 +4,750 @@
             [cljfmt.core :refer [reformat-string default-line-separator
                                  normalize-newlines find-line-separator
                                  replace-newlines wrap-normalize-newlines]]
-            [clojure.string :as str]))
+            [cljfmt.test-util.common :as common])
+  #?(:clj (:require [cljfmt.test-util.clojure])
```
weavejester: Rather than having two `:require` statements, it would be better just to have one.

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -4,289 +4,750 @@
...
-  (is (= (reformat-string "(juxt +' -')") "(juxt +' -')"))
-  (is (= (reformat-string "#\"(?i)foo\"") "#\"(?i)foo\""))
-  (is (= (reformat-string "#\"a\nb\"") "#\"a\nb\"")))
+  (is (reformats-to?
+       [";foo"]
+       [";foo"]))
+  (is (reformats-to?
+       ["::foo"]
+       ["::foo"])
+      "auto-resolved keyword")
```
weavejester: I don't think we need messages where the reformatting is obvious or trivial.

cljfmt/test/cljfmt/test_util/common.cljc
lread: I see your point and realize I don't feel too strongly about my preference.

cljfmt/test/cljfmt/test_util/common.cljc
lread: I didn't see it clearly documented, but now see it in user examples. Will comply.

cljfmt/test/cljfmt/core_test.cljc
lread: I am certainly a newbie in some areas of Clojure, if I looked it up to double check what it is, I probably added a msg.  I'll review.

weavejester:
Thanks, this looks all good. Can you squash your commits down? Then I should be able to merge.

lread:
squash in progress...

lread:
Done. Maybe. Please let me know if I've squashed to expectations.  Thanks.

weavejester:
I'm not quite sure what you've done, but you've changed several existing commits and added in a superfluous merge commit. Try running the following commands:

    git reset --hard origin/master
    git cherry-pick --strategy=recursive -Xtheirs 28a13c8d4bb8215dba6b3c40e575cda9a16cb82f
    git push -f

lread:
Oh sorry about that, my rebasing is obviosuly very very rusty.

lread:
Thanks, you saved me!

weavejester:
No problem. I think there's just one more thing to do, which is to adjust the commit message slightly. You can change it with:

    git commit --amend

Change it to:

```
Refactor reformat tests to be more readable

Adds a 'reformats-to?' custom assertion. Fixes #145.
```

I think everything else is evident from the diff.

lread:
Ok by me. Done.

PR #153, opened 2018-12-02
Fix inner indentation with index

### LINKED ISSUE #73: Odd indentation within letfn forms.
Presently, `[cljfmt "0.5.3"]` takes the following form (which I consider to be indented correctly):

``` clojure
(letfn [(foo [x]
          (* x x))
        (bar [x]
          (+ (foo x) (foo x)))]
  (bar 5))
```

and suggests/fixes the indentation to:

``` clojure
(letfn [(foo [x]
          (* x x))
        (bar [x]
             (+ (foo x) (foo x)))]
  (bar 5))
```

ProjectFrank:
`idx` -> `(inc idx)`

According to the README, the index specified for inner indentation is
the index of the argument.

For example, specifying an index of 0 means that the inner indentation
is targeting the first argument. However, the first argument does not
have an index of 0 within the form. It has an index of 1, because the
first argument should be the second expression in the current list.

`(dec depth)` -> `depth`

Considering that `(first (iterate z/up zloc))` is `zloc` itself, it is
not necessary to decrement `depth`.

Fixes #73.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -211,7 +211,7 @@
 
 (defn- index-matches-top-argument? [zloc depth idx]
   (and (> depth 0)
-       (= idx (index-of (nth (iterate z/up zloc) (dec depth))))))
+       (= (inc idx) (index-of (nth (iterate z/up zloc) depth)))))
 
 (defn- fully-qualify-symbol [possible-sym alias-map]
   (if-let [ns-string (and (symbol? possible-sym)
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -132,6 +132,16 @@
           "          (* x x))]"
           "  (foo 5))"]))
     (is (reformats-to?
+         ["(letfn [(foo [x]"
+          "(* x x))"
+          "(bar [x]\n(+ x x))]"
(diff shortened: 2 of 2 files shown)


weavejester:
Thanks for the PR! Due to a refactor of the tests (#146), you'll unfortunately have to resolve the conflict that introduced before this commit can be merged.

ProjectFrank:
@weavejester done!

weavejester:
Thanks! Can you remove the backticks from the commit message? Those are used in markdown, but commit messages are plaintext, rather than markdown, so the markup is unnecessary. If you need to denote the difference between a function and a word, you can use single quotes.

So for example:

    Fix inner indentation with index

    idx -> (inc idx)

    According to the README, the index specified for inner indentation is
    the index of the argument.

    For example, specifying an index of 0 means that the inner indentation
    is targeting the first argument. However, the first argument does not
    have an index of 0 within the form. It has an index of 1, because the
    first argument should be the second expression in the current list.

    (dec depth) -> depth

    Considering that '(first (iterate z/up zloc))' is 'zloc' itself, it is
    not necessary to decrement 'depth'.

    Fixes #73.

ProjectFrank:
@weavejester Sorry, I was being lazy and taking advantage of how GitHub uses your commit message as a PR description if your PR only has one commit when you open it. I've updated the commit message just now.

PR #156, opened 2018-12-09
Fix indentation of closing parens for empty blocks

### LINKED ISSUE #111: Parens after trailing comment not indented

### LINKED ISSUE #152: Some ”empty” forms get inconsistent formatting when keeping surrounding whitespace
When indenting and keeping surrounding whitespace some forms, like `comment` and `let` get the wrong formatting if they are ”empty” and have newlines. Like this:

```clojure
  (cljfmt/reformat-string "(let []\n)"
                          {:remove-surrounding-whitespace? false
                           :indentation? true}) => "(let []\n     )"
```

I.e. the closing paren is aligned with the first argument of the form. Which is not how let is fomatted otherwise. It should have two space indents for the code below the bindings box.

For empty `comment` forms it is that the closing paren gets one indent when it should have two, like it has when it has contents. I think I have noticed this on other forms too.

For Calva Formatter this is important because it is used while typing and many forms start out empty.

lread:
Fixes #111 and fixes #152.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -236,10 +236,15 @@
       (or (zlinebreak? zloc) (comment? zloc)))
     true))
 
+(defn- codeless-block? [zloc]
+  (nil? (z/right zloc)))
+
 (defn- block-indent [zloc key idx alias-map]
   (if (or (indent-matches? key (fully-qualify-symbol (form-symbol zloc) alias-map))
           (indent-matches? key (remove-namespace (form-symbol zloc))))
-    (if (and (some-> zloc (nth-form (inc idx)) first-form-in-line?)
+    (if (and (or
+              (codeless-block? zloc)
+              (some-> zloc (nth-form (inc idx)) first-form-in-line?))
              (> (index-of zloc) idx))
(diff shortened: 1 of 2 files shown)


weavejester:
Thanks for the fix!

Issue #125, opened 2018-06-24
Not able to format source code with #:: in it


olecve:
Hi!

lein-cljfmt 0.5.7 fails on following source code

```clojure
(ns ui-app.re-frame.db)

(def default-db #::{:page :home})
```

my project dependencies are

```clojure
[org.clojure/clojure "1.9.0"]
[org.clojure/clojurescript "1.10.238" :scope "provided"]
[lein-cljfmt "0.5.7"]
```

I tried to exclude rewrite-clj and use [rewrite-clj "0.6.0"], but it doesn't help

danielcompton:
@olecve can you share how it failed? What was the stacktrace?

olecve:
@danielcompton please find the log with stacktrace below

```
Reformatting src/cljs/ui_app/activities_view/core.cljs
Failed to format file: src/cljs/ui_app/re_frame/db.cljs
clojure.lang.ExceptionInfo: Invalid token: :
 at clojure.core$ex_info.invokeStatic (core.clj:4617)
    clojure.core$ex_info.invoke (core.clj:4617)
    clojure.tools.reader.reader_types$reader_error.invokeStatic (reader_types.clj:391)
    clojure.tools.reader.reader_types$reader_error.doInvoke (reader_types.clj:387)
    clojure.lang.RestFn.invoke (RestFn.java:439)
    clojure.tools.reader.edn$read_keyword.invokeStatic (edn.clj:269)
    clojure.tools.reader.edn$read_keyword.invoke (edn.clj:257)
    clojure.tools.reader.edn$read.invokeStatic (edn.clj:369)
    clojure.tools.reader.edn$read.invoke (edn.clj:337)
    clojure.tools.reader.edn$read.invokeStatic (edn.clj:358)
    clojure.tools.reader.edn$read.invoke (edn.clj:337)
    clojure.tools.reader.edn$read.invokeStatic (edn.clj:355)
    clojure.tools.reader.edn$read.invoke (edn.clj:337)
    rewrite_clj.parser.keyword$parse_keyword.invokeStatic (keyword.clj:14)
    rewrite_clj.parser.keyword$parse_keyword.invoke (keyword.clj:8)
    rewrite_clj.parser.core$eval2753$fn__2754.invoke (core.clj:95)
    clojure.lang.MultiFn.invoke (MultiFn.java:229)
    rewrite_clj.reader$read_with_meta.invokeStatic (reader.clj:132)
[...]

PR #158, opened 2018-12-10
Add tests for keyword and alias namespaced maps

lread:
Relates to #125.

### DIFF
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -842,6 +842,12 @@
        ["::foo/bar"]
        ["::foo/bar"]))
   (is (reformats-to?
+       ["#:x {:a :b :c :d}"]
+       ["#:x {:a :b :c :d}"]))
+  (is (reformats-to?
+       ["#::x {:a :b :c :d}"]
+       ["#::x {:a :b :c :d}"]))
+  (is (reformats-to?
        ["foo:bar"]
        ["foo:bar"])
       "a name with an embedded colon")


lread:
Hmmm... good question. I don't see any specific advice in the [style guide](https://github.com/bbatsov/clojure-style-guide).  
[Clojure's own examples](https://clojure.org/guides/weird_characters#__code_code_and_code_code_namespace_map_syntax) show the `#:x{:a 1}` formatting. I assume that's where folks will learn about this feature and probably adopt the example syntax therein. 
Soo... maybe I need to tweak this one.

danielcompton:
The Clojure printer prints it as `#:x{:a 1}`, so I'd say that is probably the better route to go down.

weavejester:
Let's go with `#:x{:a 1}`, then, as the idiomatic formatting.

weavejester:
I think so, though with the caveat that we add an option to turn the behavior off for those that want it.

lread:
An option, because it needs a name, leads me to think about generalizing.

I wonder how many other similar syntaxes would fit into this category and what their current behaviors are. For examples:
1. Cljfmt  will reformat ``#_   (+ 1 2)`` to ``#_(+ 1 2)`` if `:remove-surrounding-whitespace` is `true`.
2. But, `#?      (:clj(+ 1 2))` is not affected.

lread:
I should probably understand what these # syntaxes are commonly called.  

In our usage here it seems [# is the dispatch character](https://clojure.org/guides/weird_characters#dispatch).

So would we have `:remove-whitespace-after-dispatch?`?

(the focus of this pull request has totally changed from my original intent, I'll rename if we decide to continue - or if you prefer we could reject and move this to a new issue)

weavejester:
That seems reasonable. Feel free to open a new PR and reference this one.

PR #160, opened 2018-12-13
Fix indent for [[:block 0]] forms when the close paren is on a new line

### LINKED ISSUE #159: cond clauses formats differently when incomplete than when complete
These two should be indented the same (and like the second one):

```clojure
(cond foo
  )

(cond foo
      bar)
```

But they get indented as shown. Here's a repro:

```clojure
(cljfmt/reformat-string "(cond foo\n)\n\n(cond foo\nbar)"
                        {:remove-surrounding-whitespace? false})
=> "(cond foo\n  )\n\n(cond foo\n      bar)"
```

Not sure if this was introduced by #156, but it seems related at least.

mseddon:
Special cases [[:block 0]] forms, so that #156 doesn't interfere with their formatting.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -247,7 +247,7 @@
   (if (or (indent-matches? key (fully-qualify-symbol (form-symbol zloc) alias-map))
           (indent-matches? key (remove-namespace (form-symbol zloc))))
     (if (and (or
-              (codeless-block? zloc)
+              (and (codeless-block? zloc) (not= idx 0))
               (some-> zloc (nth-form (inc idx)) first-form-in-line?))
              (> (index-of zloc) idx))
       (inner-indent zloc key 0 nil alias-map)
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -63,7 +63,7 @@
          ["(cond->> x"
           "  a? a"
           "  b? b)"])))
-
+  
   (testing "constant indentation"
     (is (reformats-to?
          ["(def foo"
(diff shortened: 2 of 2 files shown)


weavejester:
Can you format your commit message based on these [seven rules](https://chris.beams.io/posts/git-commit/#seven-rules)?

Could you also give an explanation of why this issue occurs, and why we need a special case?

mseddon:
This PR narrows the applicability of the special case introduced in #156. In that PR, a test was added: https://github.com/weavejester/cljfmt/pull/156/commits/2a93f2c2a997d1846f626c87adabe59fad2ae9fc#diff-73e83bd1434934866cef2d6e68e43203R245

This is a good fix for the case of `let`, where we want:

```clojure
(let [x]
   )
```
The test checks to see if we're at the end of the form, and forcefully overrides `block-indent` to be an `inner-indent`.

However, #156 introduces new strangeness, specifically for formatters that use `[[:block 0]]` when using `{:remove-surrounding-whitespace? false}`:

```clojure
(alt! foo
   )
```
```clojure
(cond foo
   )
```

This specifically matters in `[[:block 0]]` forms because we want all the children (and the closing paren) to line up.

So by disabling this feature in this case allows #156 to continue functioning, but also reverts back the original behaviour in cases that were not previously problematic.

mseddon:
*ahem* sorry about that. That should pass now.

weavejester:
I've taken a look at this in more detail, and I don't think we need an exception here. Rather, the commit in #156 introduced a condition that was overly broad, so I've written a fix for it, and I pulled your tests into my commit (I hope that's okay).

mseddon:
No problem! Really my main concern was we have issues downstream into `calva-fmt`, so however it gets fixed is fine by me.

PR #167, opened 2019-03-25
Fix for "Qualifed indent specs are not applied in the referred namespace itself"

### LINKED ISSUE #166: Qualifed indent specs are not applied in the referred namespace itself
Hi there,

I found an issue which is best explained with a suggested failing test (which I ran and indeed fails):

```clojure
    (is (reformats-to?
         ["(ns thing.core"
          ")"
          ""
          "(defthing foo [x]"
          "(+ x 1))"]
         ["(ns thing.core"
          ""
          "(defthing foo [x]"
          "  (+ x 1))"]
         {:indents {'thing.core/defthing [[:inner 0]]}})
        "recognises the current namespace as part of a qualifed indent spec")
```

i.e., if `*ns*` is `thing.core`, then cljfmt should look for an unqualified `defthing` (in addition to the current behavior).

I can give a shot to a fix, but perhaps you have it easy. Let me know

Cheers - Victor

vemv:
Fixes #166.

---

Old version (i.e. first commit):

Merely a (working) POC with some acknowledged defects:

* Uses `clojure.tools.namespace` for ns parsing, I guess a rewrite-clj implementation is a must for cljs compat.
* A dynamic var is introduced. I recall you don't favor them. OTOH `defn- inner-indent [zloc key depth idx alias-map]` (and several other defns) would get another argument, which might become unwieldy.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -7,7 +7,7 @@
                  [rewrite-clj.node :as n]
                  [rewrite-clj.parser :as p]
                  [rewrite-clj.zip :as z
-                  :refer [append-space edn skip whitespace-or-comment?]])
+                  :refer [append-space edn skip whitespace-or-comment? child-sexprs]])
        (:import java.util.regex.Pattern)]
       :cljs
       [(:require [cljs.reader :as reader]
@@ -16,7 +16,7 @@
                  [rewrite-clj.node :as n]
                  [rewrite-clj.parser :as p]
                  [rewrite-clj.zip :as z]
-                 [rewrite-clj.zip.base :as zb :refer [edn]]
+                 [rewrite-clj.zip.base :as zb :refer [edn child-sexprs]]
                  [rewrite-clj.zip.whitespace :as zw
                   :refer [append-space skip whitespace-or-comment?]])
        (:require-macros [cljfmt.core :refer [read-resource]])]))
@@ -55,8 +55,14 @@
   (and (p? zloc) (or (nil? (zip/left zloc))
                      (nil? (skip zip/right p? zloc)))))
 
+(defn root? [zloc]
+  (nil? (zip/up zloc)))
+
 (defn- top? [zloc]
   (and zloc (not= (z/node zloc) (z/root zloc))))
+
+(defn- top [zloc]
+  (if (root? zloc) zloc (recur (z/up zloc))))
 
 (defn- surrounding-whitespace? [zloc]
   (and (top? (z/up zloc))
@@ -192,6 +198,18 @@
 (defn pattern? [v]
   (instance? #?(:clj Pattern :cljs js/RegExp) v))
 
+(defn- top-level-form [zloc]
+  (->> zloc
+       (iterate z/up)
+       (take-while (complement root?))
+       last))
+
+(defn- ns-form? [zloc]
+  (some-> zloc child-sexprs first (= 'ns)))
+
+(defn- find-namespace [zloc]
+  (some-> zloc top (z/find z/next ns-form?) z/sexpr second))
+
 (defn- indent-matches? [key sym]
   (cond
     (symbol? key) (= key sym)
@@ -213,16 +231,24 @@
   (and (> depth 0)
        (= (inc idx) (index-of (nth (iterate z/up zloc) depth)))))
 
-(defn- fully-qualify-symbol [possible-sym alias-map]
-  (if-let [ns-string (and (symbol? possible-sym)
-                          (namespace possible-sym))]
-    (symbol (get alias-map ns-string ns-string)
-            (name possible-sym))
-    possible-sym))
+(defn- qualify-symbol-by-alias-map [possible-sym alias-map]
+  (when-let [ns-str (namespace possible-sym)]
+    (symbol (get alias-map ns-str ns-str) (name possible-sym))))
+
+(defn- qualify-symbol-by-ns-form [possible-sym zloc]
+  (when-let [ns-name (find-namespace zloc)]
+    (symbol (name ns-name) (name possible-sym))))
+
+(defn- fully-qualified-symbol [zloc alias-map]
+  (let [possible-sym (form-symbol zloc)]
+    (if (symbol? possible-sym)
+      (or (qualify-symbol-by-alias-map possible-sym alias-map)
+          (qualify-symbol-by-ns-form possible-sym zloc))
+      possible-sym)))
 
 (defn- inner-indent [zloc key depth idx alias-map]
   (let [top (nth (iterate z/up zloc) depth)]
-    (if (and (or (indent-matches? key (fully-qualify-symbol (form-symbol top) alias-map))
+    (if (and (or (indent-matches? key (fully-qualified-symbol zloc alias-map))
                  (indent-matches? key (remove-namespace (form-symbol top))))
              (or (nil? idx) (index-matches-top-argument? zloc depth idx)))
       (let [zup (z/up zloc)]
@@ -242,7 +268,7 @@
          true)))
 
 (defn- block-indent [zloc key idx alias-map]
-  (if (or (indent-matches? key (fully-qualify-symbol (form-symbol zloc) alias-map))
+  (if (or (indent-matches? key (fully-qualified-symbol zloc alias-map))
           (indent-matches? key (remove-namespace (form-symbol zloc))))
     (let [zloc-after-idx (some-> zloc (nth-form (inc idx)))]
       (if (and (or (nil? zloc-after-idx) (first-form-in-line? zloc-after-idx))
@@ -313,9 +339,6 @@
   ([form indents alias-map]
    (indent (unindent form) indents alias-map)))
 
-(defn root? [zloc]
-  (nil? (zip/up zloc)))
-
 (defn final? [zloc]
   (and (nil? (zip/right zloc)) (root? (zip/up zloc))))
 
@@ -343,20 +366,10 @@
        (cond-> (:remove-trailing-whitespace? opts true)
          remove-trailing-whitespace))))
 
-(defn- top-level-form [zloc]
-  (->> zloc
-       (iterate z/up)
-       (take-while (complement root?))
-       last))
-
 #?(:clj
    (defn- ns-require-form? [zloc]
-     (and (some-> zloc
-                  top-level-form
-                  z/child-sexprs
-                  first
-                  (= 'ns))
-          (some-> zloc z/child-sexprs first (= :require)))))
+     (and (some-> zloc top-level-form ns-form?)
+          (some-> zloc child-sexprs first (= :require)))))
 
 #?(:clj
    (defn- as-keyword? [zloc]
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -241,6 +241,20 @@
           #?@(:cljs [:alias-map {"t" "thing.core"}])})
         "applies custom indentation to namespaced defn")
     (is (reformats-to?
+         ["(comment)"
+          "(ns thing.core)"
+          ""
+          "(defthing foo [x]"
+          "(+ x 1))"]
+         ["(comment)"
+          "(ns thing.core)"
+          ""
+          "(defthing foo [x]"
+          "  (+ x 1))"]
+         {:indents {'thing.core/defthing [[:inner 0]]}
+          #?@(:cljs [:alias-map {}])})
+        "recognises the current namespace as part of a qualifed indent spec")
+    (is (reformats-to?
          ["(ns example"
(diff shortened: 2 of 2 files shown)


weavejester:
Thanks for the PR. I think we can avoid dynamic vars and a separate parsing from tools.namespace entirely. Instead, create a function to get the current namespace given a zloc, and then update `inner-indent` to pass the current namespace to `fully-qualify-symbol`.

vemv:
> Instead, create a function to get the current namespace given a zloc

Understood, this would avoid having to add an extra arg to potentially many defns.

Btw, I just realised there's another issue to tackle: `:refer [x y z]` clauses are not considered.

e.g., if a ns refers `foo` from `another-ns`, there's no attempted resolution against a `another-ns/foo` indent spec which might exist.

Should include that fix (if achieved) in the same PR?

weavejester:
I think support of `:refer` can be tackled in a separate PR.

vemv:
Ready now :)

The only new function is ns-name-from-zloc (the rest being mostly moved-around items). Let me now if it should have explicit unit testing.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -192,10 +192,37 @@
...
+       (take-while (complement root?))
+       last))
+
+(defn- ns-form? [zloc]
+  (some-> zloc
+          child-sexprs
+          first
+          #{'ns `ns}))
+
+(defn- ns-name-from-zloc [zloc]
```
weavejester: Better name would be `find-namespace`. Function should return a symbol. I don't know why you have `edn` or have that `if` in there.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -192,10 +192,37 @@
...
+  (let [node (some-> zloc z/root edn)]
+    (if (ns-form? node)
+      node
+      (some-> node (z/find-next ns-form?)))))
+
+(defn- indent-matches? [zloc key sym]
```
weavejester: We should change `fully-qualify-symbol`, not `indent-matches?`.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -7,7 +7,7 @@
                  [rewrite-clj.node :as n]
                  [rewrite-clj.parser :as p]
                  [rewrite-clj.zip :as z
-                  :refer [append-space edn skip whitespace-or-comment?]])
+                  :refer [append-space edn skip whitespace-or-comment? child-sexprs]])
```
weavejester: Why did you add `child-sexprs`?

cljfmt/src/cljfmt/core.cljc
vemv: Good one!

vemv:
Ready again!

Force-pushed due to some bad formatting.

vemv:
Reminder

cljfmt/src/cljfmt/core.cljc
```diff
@@ -192,6 +192,28 @@
...
+       (take-while (complement root?))
+       last))
+
+(defn- ns-form? [zloc]
+  (some-> zloc
+          child-sexprs
+          first
+          #{'ns `ns}))
+
+(defn- find-namespace [zloc]
```
weavejester: A more concise way of writing this would be:

```clojure
(defn- top [zloc]
  (if (root? zloc) zloc (recur (z/up zloc))))

(defn- find-namespace [zloc]
  (some-> zloc top (z/find z/next ns-form?) z/sexpr second))
```

cljfmt/src/cljfmt/core.cljc
```diff
@@ -213,17 +235,27 @@
   (and (> depth 0)
        (= (inc idx) (index-of (nth (iterate z/up zloc) depth)))))
 
-(defn- fully-qualify-symbol [possible-sym alias-map]
+(def ^:private named? (some-fn string? keyword? symbol?))
+
+(defn- fully-qualify-symbol [possible-sym zloc alias-map]
```
weavejester: This should likely be just: `[zloc alias-map]`.

cljfmt/src/cljfmt/core.cljc
vemv: Done

cljfmt/src/cljfmt/core.cljc
vemv: Sorry, I can't follow. The `fully-qualify-symbol` naming suggest the defn is about processing a symbol.

In any case I can't imagine what kind of change is expected...

cljfmt/src/cljfmt/core.cljc
```diff
@@ -213,17 +235,27 @@
...
   (if-let [ns-string (and (symbol? possible-sym)
                           (namespace possible-sym))]
     (symbol (get alias-map ns-string ns-string)
             (name possible-sym))
-    possible-sym))
+    (if-let [ns-name (find-namespace zloc)]
+      (if (named? possible-sym)
```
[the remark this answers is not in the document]
vemv: 👍 removed, suite still passes. Probably I was dragging old habits rather than tackling an specific problem.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -192,6 +192,28 @@
...
+  (->> zloc
+       (iterate z/up)
+       (take-while (complement root?))
+       last))
+
+(defn- ns-form? [zloc]
+  (some-> zloc
+          child-sexprs
+          first
+          #{'ns `ns}))
```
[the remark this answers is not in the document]
weavejester: Maybe, except \`ns is going to be `cljfmt.core/ns`.

Let's keep it as `'ns` for now. If it turns out to be a problem, we can fix it in future. I think there's quite a few tools already that assume `ns` rather than `clojure.core/ns`.

cljfmt/src/cljfmt/core.cljc
weavejester: In all instances of `fully-qualify-symbol`, we're writing:

```clojure
(fully-qualify-symbol (form-symbol zloc) zloc alias-map)
```

Instead let's just pass the zloc of the symbol:

```clojure
(fully-qualified-symbol zloc alias-map)
```

cljfmt/src/cljfmt/core.cljc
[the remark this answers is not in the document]
vemv: It actually expands to `clojure.core/ns`

Can change, just confirm

cljfmt/src/cljfmt/core.cljc
vemv: One of the occurrences is like this:

```clojure
                     (fully-qualify-symbol (form-symbol top)
                                                        zloc
                                                        alias-map)
```

`top` being different than `zloc`

weavejester: Yeah, but it could easily just be rewritten using `top` in both locations:

```clojure
(fully-qualify-symbol (form-symbol top) top alias-map)
```

Unless I'm wrong (again!), it doesn't matter where the zloc argument is, because we're always going up to the top level.

vemv: In:

```clojure
(defn- inner-indent [zloc key depth idx alias-map]
  (let [top (nth (iterate z/up zloc) depth)]
    (if (and (or (indent-matches? key
                                  (fully-qualify-symbol (form-symbol top)
                                                        zloc
                                                        alias-map))
                 (indent-matches? key
                                  (remove-namespace (form-symbol top))))
             (or (nil? idx) (index-matches-top-argument? zloc depth idx)))
      (let [zup (z/up zloc)]
        (+ (margin zup) (indent-width zup))))))
```

...the thing called `top` is not the absolute top, but the thing up `depth` levels.

`fully-qualify-symbol` cannot guess that depth.

weavejester: It doesn't need to guess it. `find-namespace` will work with any `zloc`. So:

```clojure
(fully-qualify-symbol (form-symbol top) zloc alias-map)
```

Will produce the same result as:

```clojure
(fully-qualify-symbol (form-symbol top) top alias-map)
```

vemv: Got it! Done.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -213,17 +234,24 @@
   (and (> depth 0)
        (= (inc idx) (index-of (nth (iterate z/up zloc) depth)))))
 
-(defn- fully-qualify-symbol [possible-sym alias-map]
-  (if-let [ns-string (and (symbol? possible-sym)
-                          (namespace possible-sym))]
-    (symbol (get alias-map ns-string ns-string)
-            (name possible-sym))
-    possible-sym))
+(defn- fully-qualified-symbol [zloc alias-map]
```
weavejester: What about writing it instead as:

```clojure
(defn- qualify-symbol-by-alias-map [possible-sym alias-map]
  (when-let [ns-string (namespace possible-sym)]
    (symbol (get alias-map ns-string ns-string)
            (name possible-sym))))

(defn- qualify-symbol-by-ns-form [possible-sym zloc]
  (when-let [ns-name (find-namespace zloc)]
    (symbol (name ns-name) (name possible-sym))))

(defn- fully-qualified-symbol [zloc alias-map]
  (let [possible-sym (form-symbol zloc)]
    (or (if-not (symbol? possible-sym) possible-sym)
        (qualify-symbol-by-alias-map possible-sym alias-map)
        (qualify-symbol-by-ns-form possible-sym zloc))
        possible-sym)))
```

I think that makes the cascading logic a little clearer.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -213,17 +234,24 @@
...
+          (symbol (name ns-name) (name possible-sym))
+          possible-sym)
+        possible-sym))))
 
 (defn- inner-indent [zloc key depth idx alias-map]
   (let [top (nth (iterate z/up zloc) depth)]
-    (if (and (or (indent-matches? key (fully-qualify-symbol (form-symbol top) alias-map))
-                 (indent-matches? key (remove-namespace (form-symbol top))))
+    (if (and (or (indent-matches? key
+                                  (fully-qualified-symbol zloc alias-map))
```
weavejester: We don't need to change the formatting here.

cljfmt/src/cljfmt/core.cljc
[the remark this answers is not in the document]
weavejester: Let's change it to `(= 'ns)` for now, because this is technically a change outside the current PR. However, I'd accept a small additional PR to change it to:

     #{'ns `ns}

Also we can put it all on one line:

```clojure
(defn- ns-form? [zloc]
  (some-> zloc z/sexpr first (= 'ns)))
```

vemv:
Alright! Done.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -369,15 +382,15 @@
...
+                               (first (child-sexprs grandparent-node)))]
        (when (and (symbol? alias) (symbol? current-namespace))
          {(str alias) (if parent-namespace
                         (format "%s.%s" parent-namespace current-namespace)
                         (str current-namespace))}))))
 
 #?(:clj
    (defn- alias-map-for-form [form]
-     (when-let [require-zloc (-> form z/edn (z/find z/next ns-require-form?))]
+     (when-let [require-zloc (-> form edn (z/find z/next ns-require-form?))]
```
weavejester: This line doesn't need to be changed, as it's Clojure only.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -369,15 +382,15 @@
            current-namespace (some-> as-zloc z/leftmost z/sexpr)
            grandparent-node  (some-> as-zloc z/up z/up)
            parent-namespace  (when-not (ns-require-form? grandparent-node)
-                               (first (z/child-sexprs grandparent-node)))]
+                               (first (child-sexprs grandparent-node)))]
```
weavejester: This line doesn't need to be changed, as it's Clojure only.

vemv: OTOH, with this, all JVM clj code will consistently use the same symbols. Using `child-sexprs` in some places but `z/child-sexprs` in others can be confusing, leading one to think there's a reason for that.

Same for `edn`

Can change, just confirm

weavejester: That's not a bad idea, but it's a job for another PR. We want the diff to show changes only relevant to the current PR.

weavejester:
Okay, that all looks good! :+1:

Can you squash your changes? Then I'll merge and release.

vemv:
🍻thanks!

weavejester:
Thanks for the update! I think the commit message is slightly ambiguous, as it doesn't mention *why* the taking the current namespace into account matters. So perhaps instead:

    Account for current ns when finding indents

    Fixes #166.

PR #169, opened 2019-05-20
Add a Clojars badge

camsaul:
Saves people a few clicks by not having to go to the Releases page to check what the latest release is


weavejester:
The latest version is at the top of the README anyway. I don't think we need a badge as well.

PR #183, opened 2019-12-05
Add more guidance to indentation rules docs

### LINKED ISSUE #182: Provide more details on identation rules in README
Don't get me wrong, cljfmt is great, but every time I come back to it, I end up scratching my head when reading over how indentation rules work in the README. I expect I am not the only one who gets confused?

1. I find the concept of `:inner` and `:block` straightforward but it often takes me a while to figure out what their arguments actually mean.  So a little more hand-holding there would be nice.
1. go over the formatted code examples and verify they match today's reality.
1. try to make the section more clear in general, this might involve a bit of re-ordering of examples.
1. I understand you appreciate succinctness and will do my best to comply.

I will, unless you have strong objections, follow up with a PR for review.

lread:
Closes #182


lread:
Hi @weavejester, I took many passes at this before submitting.  I feel I've got it to a point where it is something I would have appreciated when first coming to cljfmt, and will appreciate when returning after an absence.  

All formatted code blocks were tested at the REPL.

I look forward to your review and feedback.

weavejester:
Thanks for the patch. I like the extra examples and explanations, but the table with depth etc. seems to me to be more confusing than clarifying. It took me a few minutes before I understood what it was explaining.

lread:
Hmm... interesting, thanks for the feedback.  

I think depth needs explaining.  Somehow.  Maybe a diagram would be a better vehicle.  Would you be opposed to an attempt with a diagram?  I am finding the open source [draw.io](https://about.draw.io/about-us/) to be a great drawing tool for open source projects.

Or maybe I just need to get rid of the table and express with words and code snippits.  I'll take a stab at that first.

weavejester:
I think a diagram might be a good idea. However, be aware that I don't know how long this syntax will be around for, so I'd caution you against spending too much time on it. I don't have much time to spare, so it'll likely be like this for a while, but the eventual goal is to move to Cider's syntax, so that there's a common syntax format.

lread:
Thanks @weavejester, that's helpful.  

I'll work to enlighten more than I befuddle!  A diagram just might do the trick.

lread:
I think the diagrams are an improvement. When you have a moment, looking forward to your thoughts.  

Note that that I chose [colorblind friendly colors from IBM Design Library palette as described here](https://davidmathlogic.com/colorblind/#%23648FFF-%23785EF0-%23DC267F-%23FE6100-%23FFB000).

weavejester:
Thanks, this looks good. Can you squash your commits down?

lread:
Thanks @weavejester, commits squashed.

lread:
Hi @weavejester, is there anything more I should do on this?  I suppose I could add some unit tests that test the documented examples.

weavejester:
The commit seems fine. Apologies for the delay; it's been a little busy.

lread:
Thanks @weavejester.

PR #185, opened 2019-12-23
Use pmap to parallelize check and fix

dbasedow:
Same idea as PR #104 but without using core.async.

Measurements for `lein run -m cljfmt.main check src/cljfmt/core.cljc`
Before: 15.896s
After: 15.711s

Measurements for `lein run -m cljfmt.main check large_project/src`
Before: 124.275s
After: 67.317s

(All measurements are mean over 10 runs on MacBook Pro)

### DIFF
--- a/cljfmt/src/cljfmt/main.clj
+++ b/cljfmt/src/cljfmt/main.clj
@@ -94,7 +94,8 @@
     (when-not (zero? incorrect)
       (warn incorrect "file(s) formatted incorrectly"))
     (when (and (zero? incorrect) (zero? error))
-      (println "All source files formatted correctly"))))
+      (println "All source files formatted correctly"))
+    counts))
 
 (defn- merge-counts
   ([]    zero-counts)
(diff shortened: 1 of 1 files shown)


vemv:
Hi, for a bit of context I'm the author of https://github.com/nedap/formatting-stack which was shared yesterday.

formatting-stack makes cljfmt parallel in a more efficient way than pmap.

Anyway, my concern isthat this PR would hardcode a pmap to cljfmt's API. So formatting-stack would run a pmap in parallel - which makes no sense.

I think it's better to offer APIs as simple, agnostic functions, where parallelism is a choice of the consumer.

Maybe the PR can be tweaked to move the `pmap` from `main.clj` to the Lein side?

weavejester:
A better design might be to split out the inner loop of `check` and `fix` into `check-files` and `fix-file` respectively, and then perhaps we either want an option on `check` and `fix` to optionally run it in parallel.

This PR also has a few other things that need to be fixed. The commit message needs to adhere to the contributing guidelines, `print-final-count` should not be changed (use `doto` instead), and at least one of the pmaps has an inner `println`.

PR #186, opened 2020-01-13
Native image

bsless:
Hello,

I found myself needing to build a native-image from cljfmt for my poor coworkers, so while I hacked together some side repo to help them, I thought I'd open a bit more detailed PR to facilitate it in the project itself.

Contents of this PR:
- Adds type hints to functions which broke under native-image reflection.
- Pulls in functions from `stacktrace` to add type hints to them when needed for the same reason.
- Adds stub lein app to generate a native-image from, with a compile script for GraalVM.

Please let me know if there are any issues with it or if I made some mess of the licensing

Cheers,
Ben

### DIFF
--- a/cljfmt/project.clj
+++ b/cljfmt/project.clj
@@ -10,7 +10,8 @@
                  [com.googlecode.java-diff-utils/diffutils "1.3.0"]
                  [rewrite-clj "0.6.1"]
                  [rewrite-cljs "0.4.4"]]
-  :plugins [[lein-cljsbuild "1.1.7"]]
+  :plugins [[lein-cljsbuild "1.1.7"]
+            [io.taylorwood/lein-native-image "0.3.1"]]
   :hooks [leiningen.cljsbuild]
   :cljsbuild {:builds
               {"dev" {:source-paths ["src" "test"]
@@ -21,5 +22,24 @@
                                  :optimizations :none}}}
               :test-commands
               {"dev" ["node" "target/out/tests.js"]}}
+  :native-image
+  {:name "cljfmt"
+   :opts ["--verbose"
+          "-H:+ReportExceptionStackTraces"
+          "-J-Dclojure.spec.skip-macros=true"
+          "-J-Dclojure.compiler.direct-linking=true"
+          "-H:ReflectionConfigurationFiles=reflection.json"
+          "--initialize-at-build-time"
+          "-H:Log=registerResource:"
+          "--verbose"
+          "--no-fallback"
+          "--no-server"
+          "-J-Xmx3g"]}
   :profiles
-  {:provided {:dependencies [[org.clojure/clojurescript "1.10.520"]]}})
+  {:uberjar
+   {:main cljfmt.main
+    :aot :all
+    :native-image
+    {:jvm-opts ["-Dclojure.compiler.direct-linking=true"
+                "-Dclojure.spec.skip-macros=true"]}}
+   :provided {:dependencies [[org.clojure/clojurescript "1.10.520"]]}})
--- a/cljfmt/src/cljfmt/diff.clj
+++ b/cljfmt/src/cljfmt/diff.clj
@@ -13,7 +13,7 @@
 
 (defn to-absolute-path [filename]
   (->> (str/split filename (re-pattern (Pattern/quote File/separator)))
-       (apply io/file)
+       ^java.io.File (apply io/file)
        .getCanonicalPath))
 
 (defn unified-diff
--- a/cljfmt/src/cljfmt/main.clj
+++ b/cljfmt/src/cljfmt/main.clj
@@ -3,10 +3,10 @@
   (:require [cljfmt.core :as cljfmt]
             [clojure.edn :as edn]
             [clojure.java.io :as io]
-            [clojure.string :as str]
-            [clojure.stacktrace :as st]
+            [cljfmt.impl.stacktrace :as st]
             [clojure.tools.cli :as cli]
-            [cljfmt.diff :as diff]))
+            [cljfmt.diff :as diff])
+  (:gen-class))
 
 (defn- abort [& msg]
   (binding [*out* *err*]
@@ -18,7 +18,7 @@
   (binding [*out* *err*]
     (apply println args)))
 
-(defn- relative-path [dir file]
+(defn- relative-path [^java.io.File dir ^java.io.File file]
   (-> (.toAbsolutePath (.toPath dir))
       (.relativize (.toAbsolutePath (.toPath file)))
       (.toString)))
--- a/cljfmt/reflection.json
+++ b/cljfmt/reflection.json
@@ -1,0 +1,9 @@
+[
+    {
+        "name": "java.lang.Class",
+        "allDeclaredConstructors": true,
+        "allPublicConstructors": true,
+        "allDeclaredMethods": true,
+        "allPublicMethods": true
+    }
+]
--- a/cljfmt/src/cljfmt/impl/stacktrace.clj
+++ b/cljfmt/src/cljfmt/impl/stacktrace.clj
@@ -1,0 +1,61 @@
+;;;   Copyright (c) Rich Hickey. All rights reserved.
+;;;   The use and distribution terms for this software are covered by the
+;;;   Eclipse Public License 1.0 (http://opensource.org/licenses/eclipse-1.0.php)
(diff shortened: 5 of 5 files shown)


weavejester:
Thanks the the PR, and apologies for missing this the first time around.

This looks good, but is there a reason for putting it in a separate project rather than making it a profile of the existing one? Likewise, would it be better to use something like [lein-native-image](https://github.com/taylorwood/lein-native-image) instead of a custom script?

I'd also suggest taking a look at the [contributing guidelines](https://github.com/weavejester/cljfmt/blob/master/CONTRIBUTING.md) regarding the commit messages and project version change.

bsless:
- separate project: I wanted to modify the core project itself as little as possible and to not muddy up core cljfmt with things it doesn't do. Like lein uses it as a dependency, since the jar it produces isn't executable, I thought providing an application which consumes it as a single dependency would be best. Seemed both easy and simple.
- lein native image: I had some unsuccessful experiences with it while lifting kondo's script Just Worked. Also, going by leininge's own recommendation of "The first thing to do when writing a plugin is to try to accomplish what you're doing without a plugin" I thought it was the best way.
- commit messages: I tried to stick pretty close to the guide, but I don't mind fleshing them out or changing them. I tagged the commits with the branch name out of habit, too, I don't mind removing it. Let me know if I messed it up
- version change: I'll rewrite history and remove it, apologies.

I'm not attached to the decisions I made regarding the separate project and the native image. If you prefer it done another way I don't mind changing it.

bsless:
Reworded all commit messages and removed the version bump.

weavejester:
Thanks for the commit updates.

I think I'd rather this be part of the cljfmt project, because it's just a compile target, in the same way that ClojureScript is. I don't think there's a compelling reason for it to be a separate project, aside from bringing in the core stacktrace methods, and we can potentially use a profile to hide them away.

I'd also rather that GraalVM be integrated into Leiningen, rather than having a separate script. I don't mind if this uses lein-native-image or something else. Again, the reason for this is that GraalVM is a compile target, and I'd rather not have a separate script for every platform cljfmt targets.

bsless:
Alright, I'll see what I can do :)

bsless:
Well, that was easier than I expected and completely painless

bsless:
Gentle reminder :slightly_smiling_face: 
All your requests have been implemented

cljfmt/project.clj
```diff
@@ -21,5 +22,27 @@
                                  :optimizations :none}}}
               :test-commands
               {"dev" ["node" "target/out/tests.js"]}}
+  :native-image
+  {:name "cljfmt"
+   :opts
+   ["--verbose"
```
weavejester: As a minor formatting point, can you put the vector on the same line as `:opts`, e.g.

```clojure
:opts ["--verbose"
       "-H:+ReportExceptionStackTraces"
       ...]

File: cljfmt/project.clj [hunk 2]
```diff
@@ -21,5 +22,27 @@
...
+    "-H:Log=registerResource:"
+    "--verbose"
+    "--no-fallback"
+    "--no-server"
+    "-J-Xmx3g"]}
   :profiles
-  {:provided {:dependencies [[org.clojure/clojurescript "1.10.520"]]}})
+  {:uberjar
+   {:source-paths
+    ["standalone"]
```
weavejester: Can you put `["standalone"]` on the same line as `:source-paths`. Typically we shouldn't linebreak keys unless we run out of horizontal room.

cljfmt/src/cljfmt/stacktrace.clj
```diff
@@ -0,0 +1,61 @@
+;;;   Copyright (c) Rich Hickey. All rights reserved.
```
weavejester: Let's change the namespace name to something like `cljfmt.impl.stacktrace` to indicate it's not for public use.

cljfmt/src/cljfmt/stacktrace.clj
```diff
@@ -0,0 +1,61 @@
...
+
+;;; Bring over the functions from stacktrace because they need to be
+;;; type hinted for the native-image to work
+
+(defn print-trace-element
+  "Prints a Clojure-oriented view of one element in a stack trace."
+  {:added "1.1"}
+  [^java.lang.StackTraceElement e]
+  (let [class (.getClassName e)
+	      method (.getMethodName e)
```
weavejester: Indentation is off here.

cljfmt/src/cljfmt/stacktrace.clj
```diff
@@ -0,0 +1,61 @@
...
+        match (re-matches #"^([A-Za-z0-9_.-]+)\$(\w+)__\d+$" (str class))]
+    (if (and match (= "invoke" method))
+	    (apply printf "%s/%s" (rest match))
```
weavejester: Indentation is off here, too.

cljfmt/src/cljfmt/stacktrace.clj
```diff
@@ -0,0 +1,61 @@
...
+   (let [st (.getStackTrace tr)]
+     (print-throwable tr)
+     (newline)
+     (print " at ")
+     (if-let [e (first st)]
+       (print-trace-element e)
+       (print "[empty stack trace]"))
+     (newline)
+     (doseq [e (if (nil? n)
+		             (rest st)
```
weavejester: Indentation is off here. You may want to run cljfmt over this code.

cljfmt/standalone/cljfmt/standalone.clj
```diff
@@ -0,0 +1,7 @@
+(ns cljfmt.standalone
```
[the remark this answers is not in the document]
bsless: Won't it effect the artifact generated from `cljfmt.main`? It's used by the lein task as a dependency, and I don't know if it has any effects, so I wanted to make sure to keep them separate.
I don't mind trying.

cljfmt/src/cljfmt/diff.clj
```diff
@@ -12,9 +12,13 @@
   (str/join "\n" ss))
 
 (defn to-absolute-path [filename]
-  (->> (str/split filename (re-pattern (Pattern/quote File/separator)))
-       (apply io/file)
-       .getCanonicalPath))
+  (.getCanonicalPath
```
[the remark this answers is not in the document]
bsless: I didn't know writing the meta like that works but apparently it does. Will do.

cljfmt/project.clj
bsless: Sure

cljfmt/src/cljfmt/diff.clj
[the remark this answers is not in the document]
bsless: done

cljfmt/standalone/cljfmt/standalone.clj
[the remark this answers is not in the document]
bsless: done

weavejester:
This all looks fine. Can you squash your commits down?

bsless:
Squashed

weavejester:
Thanks! Can you change the commit message to:

    Add native image building with GraalVM

bsless:
Done :)

weavejester:
Thanks, but the commit body still has all of the old commit messages. Can you remove those as well, so that the commit is *just*:

    Add native image building with GraalVM

bsless:
Ah, sure!
Done :)

cljfmt/src/cljfmt/main.clj
```diff
@@ -19,8 +19,8 @@
     (apply println args)))
 
 (defn- relative-path [dir file]
```
vemv: `[^File dir ^File file]` tends to be more DRY / future-proof than the current hinting, since otherwise a future change could introdcue a new `dir` reference that is not type-hinted.

weavejester: This seems like a good change to make.

bsless: done

bsless:
Hey guys, I did the requested changes some three weeks ago, any update?

weavejester:
Sorry for leaving you hanging. Right now I unfortunately don't have the time to test this, but my schedule should free up in a week or so.

bsless:
No rush, just didn't want to see the branch rot. Hope you're staying safe, given the circumstances.

weavejester:
Merged. Sorry this took a while to get tested and approved.

PR #193, opened 2020-03-19
Lint project.clj and/or deps.edn if they exist

### LINKED ISSUE #192: Need a blank project.clj to use tools.deps version?
Apologies if this is me doing something not quite right in my projects, but, well:

Steps:  
1. Add `cljfmt {:mvn/version "0.6.7"}` to `deps.edn`
2. Run `clojure -m cljfmt.main check` in a project that has a `deps.edn` file but no `project.clj`

Expected:  
The linter should run.

Actual:  
shell reports `No such file: project.clj`

Notes:
Doing `touch project.clj` to get a completely blank Lein project file and then re-running gives me `All source files formatted correctly`

rgm:
The default paths should include deps.edn, for symmetry with #150. This
commit adds an existence check for all default paths, to avoid failing
on a missing default path. The current set should cover these common
project types:

- Leiningen projects with `project.clj` present
- tools.deps projects with `deps.edn` present
- Combinations where both are present, eg. lein projects using the
  lein-tools-deps plugin.

Resolves #192.

### DIFF
--- a/cljfmt/src/cljfmt/main.clj
+++ b/cljfmt/src/cljfmt/main.clj
@@ -191,11 +191,14 @@
 (defn- command-name []
   (or (System/getProperty "sun.java.command") "cljfmt"))
 
+(defn- file-exists? [path]
+  (.exists (io/as-file path)))
+
 (defn -main [& args]
   (let [parsed-opts   (cli/parse-opts args cli-options)
         [cmd & paths] (:arguments parsed-opts)
         options       (merge-default-options (:options parsed-opts))
-        paths         (or (seq paths) default-paths)]
+        paths         (or (seq paths) (filter file-exists? default-paths))]
     (if (:errors parsed-opts)
       (abort (:errors parsed-opts))
       (if (or (nil? cmd) (:help options))


rgm:
@weavejester This fixes the immediate issue, but I suppose there's a conversation to be had about whether we should be checking `deps.edn` by default in this code path (similar to #150).

Would we generally expect anyone running `cljfmt.main` to be doing it in a leiningen project, or should I figure out a little more logic to do the right thing, eg. just nil out if one or the other isn't present?

weavejester:
Thanks, but ideally a bugfix shouldn't remove existing functionality.

Rather than simply removing `project.clj` from the default files to check, may I suggest instead that you filter out any default path that does not exist?

Also, please ensure that your commit message adheres to the contributing guidelines.

rgm:
Hoping that meets your comments. I changed the title to reflect the symmetry with #150, and reviewed the git commit seven rules. Please let me know if there's anything else I can adjust.

cljfmt/src/cljfmt/main.clj
```diff
@@ -156,7 +156,9 @@
       (assoc :indents (merge (:indents default-options)
                              (:indents options {})))))
 
-(def default-paths ["src" "test" "project.clj"])
+(def default-paths
+  (let [exists? (fn [path] (.exists (io/as-file path)))]
```
weavejester: Can you factor this out into a private function `file-exists?`?

cljfmt/src/cljfmt/main.clj
```diff
@@ -156,7 +156,9 @@
...
+    (filter exists? ["src" "test" "project.clj" "deps.edn"])))
```
weavejester: This needs to be in`-main`, not part of the `default-paths` definition. You can't be sure that this def is evaluated at the same time as the `-main` function is executed.

rgm: Oh, interesting. So if I grok that right, we could check existence at process startup, the world shifts out from under us, and by the time our check/fix gets invoked the existence pattern might have changed. Yep, will fix.

weavejester: No, worse than that. The filter is applied when the file is compiled. With AOT, compilation may take place at an entirely different time, and on an entirely different machine, than execution.

cljfmt/src/cljfmt/main.clj
```diff
@@ -191,11 +191,14 @@
...
 
+(defn- file-exists? [path]
+  (.exists (io/as-file path)))
+
 (defn -main [& args]
   (let [parsed-opts   (cli/parse-opts args cli-options)
         [cmd & paths] (:arguments parsed-opts)
         options       (merge-default-options (:options parsed-opts))
-        paths         (or (seq paths) default-paths)]
+        paths         (filter file-exists? (or (seq paths) default-paths))]
```
weavejester: This should be:

```clojure
(or (seq paths) (filter file-exists? default-paths))
```

If people add a path manually that doesn't exist, we *do* want to throw an error.

cljfmt/src/cljfmt/main.clj
```diff
@@ -156,7 +156,7 @@
       (assoc :indents (merge (:indents default-options)
                              (:indents options {})))))
 
-(def default-paths ["src" "test" "project.clj"])
+(def default-paths ["src" "test" "project.clj" "deps.edn"])
```
weavejester: Let's keep this PR focused and remove `"deps.edn"` for now. That should be part of a separate PR.

cljfmt/src/cljfmt/main.clj
rgm: Hadn't thought of that. Thanks.

cljfmt/src/cljfmt/main.clj
rgm: Probably fine to just feed it in manually when desired; it's not quote-unquote "code" in the same sense as `project.clj`.

djblue:
This PR seems to solve my problem too. Is there anything I can do to help it get in?

weavejester:
I don't get a notification from Github when a new commit is added, only when a new comment is, so you've already helped out, @djblue. This PR looks fine - the comments just need to be squashed, ready for commit.

rgm:
@weavejester squashed and ready.

weavejester:
Thanks. Let's update the commit message as well so it adheres to the contributing guidelines, and reflects the changes to the PR. So something more like:

```
Limit default paths in -main to ones that exist
```

rgm:
Done.

PR #195, opened 2020-04-26
Add split-maps? option, break hashmaps into lines

doby162:
The clojure style guide includes several valid hashmap structures,
inline, comma separated, and newline separated. This adds explicit
support for newline separated hashmaps, so that they can be
uniformly applied across a project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,9 @@ selectively enabled or disabled:
   true if cljfmt should collapse consecutive blank lines. This will
   convert `(foo)\n\n\n(bar)` to `(foo)\n\n(bar)`. Defaults to true.
 
+* `:split-keypairs-over-multiple-lines?` -
+  true if cljfmt should break hashmaps onto multiple lines. This will
+  convert `{:a 1 :b 2}` to `{:a 1\n:b 2}`. Defaults to false.
 
 You can also configure the behavior of cljfmt:
 
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -179,8 +179,12 @@
 (defn- coll-indent [zloc]
   (-> zloc zip/leftmost margin))
 
+(defn- uneval? [zloc]
+  (= (z/tag zloc) :uneval))
+
 (defn- index-of [zloc]
   (->> (iterate z/left zloc)
+       (remove uneval?)
        (take-while identity)
        (count)
        (dec)))
@@ -341,6 +345,29 @@
   ([form indents alias-map]
    (transform form edit-all should-indent? #(indent-line % indents alias-map))))
 
+(defn- map-key? [zloc]
+  (and (z/map? (z/up zloc))
+       (even? (index-of zloc))
+       (not (uneval? zloc))
+       (not (whitespace-or-comment? zloc))))
+
+(defn- preceded-by-line-break? [zloc]
+  (loop [previous (zip/left zloc)]
+    (cond
+      (line-break? previous)
+      true
+      (whitespace-or-comment? previous)
+      (recur (zip/left previous)))))
+
+(defn- map-key-without-line-break? [zloc]
+  (and (map-key? zloc) (not (preceded-by-line-break? zloc))))
+
+(defn- insert-newline-left [zloc]
+  (zip/insert-left zloc (n/newlines 1)))
+
+(defn split-keypairs-over-multiple-lines [form]
+  (transform form edit-all map-key-without-line-break? insert-newline-left))
+
 (defn reindent
   ([form]
    (indent (unindent form)))
@@ -364,6 +391,8 @@
    (reformat-form form {}))
   ([form opts]
    (-> form
+       (cond-> (:split-keypairs-over-multiple-lines? opts false)
+         (split-keypairs-over-multiple-lines))
        (cond-> (:remove-consecutive-blank-lines? opts true)
          remove-consecutive-blank-lines)
        (cond-> (:remove-surrounding-whitespace? opts true)
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -926,7 +926,59 @@
         " bar "
         " )"]
        {:remove-surrounding-whitespace? false
-        :remove-trailing-whitespace? false})))
+        :remove-trailing-whitespace? false}))
+  (is (reformats-to?
+       ["{:one two :three four}"]
+       ["{:one two"
+        " :three four}"]
+       {:split-keypairs-over-multiple-lines? true}))
+  (is (reformats-to?
+       ["{:one two"
+        " :three four}"]
+       ["{:one two"
+        " :three four}"]
+       {:split-keypairs-over-multiple-lines? true}))
+  (is (reformats-to?
+       ["{:one two"
+        ";comment"
+        ":three four}"]
+       ["{:one two"
+        ";comment"
+        " :three four}"]
+       {:split-keypairs-over-multiple-lines? true}))
+  (is (reformats-to?
+       ["{:one two ;comment"
+        ":three four}"]
+       ["{:one two ;comment"
+        " :three four}"]
+       {:split-keypairs-over-multiple-lines? true}))
+  (is (reformats-to?
+       ["{;comment"
+        ":one two"
+        ":three four}"]
+       ["{;comment"
+        " :one two"
+        " :three four}"]
+       {:split-keypairs-over-multiple-lines? true}))
+  (is (reformats-to?
+       ["{:one two, :three four}"]
+       ["{:one two, "
+        " :three four}"]
+       {:split-keypairs-over-multiple-lines? true
+        :remove-trailing-whitespace? false}))
+  (is (reformats-to?
+       ["{:one two,"
+        " :three four}"]
+       ["{:one two,"
+        " :three four}"]
+       {:split-keypairs-over-multiple-lines? true
+        :remove-trailing-whitespace? false}))
+  (is (reformats-to?
+       ["{:one two #_comment"
+        ":three four}"]
+       ["{:one two #_comment"
+        " :three four}"]
+       {:split-keypairs-over-multiple-lines? true})))
 
 (deftest test-parsing
   (is (reformats-to?


README.md
```diff
@@ -132,6 +132,10 @@ selectively enabled or disabled:
   true if cljfmt should collapse consecutive blank lines. This will
   convert `(foo)\n\n\n(bar)` to `(foo)\n\n(bar)`. Defaults to true.
 
+* `:split-maps?` -
```
weavejester, line 135: Let's make this a little more descriptive. For example: `:split-keypairs-over-multiple-lines?`.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -341,6 +341,25 @@
   ([form indents alias-map]
    (transform form edit-all should-indent? #(indent-line % indents alias-map))))
 
+(defn split-maps
+  "Puts each hashmap pair on it's own line"
```
weavejester, line 345: These functions are for internal use, so don't require docstrings.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -341,6 +341,25 @@
...
+  [form]
+  (transform form edit-all z/map?
+             (fn [zloc]
```
weavejester, line 348: This inner function should be split up into more descriptive private functions.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -341,6 +341,25 @@
   ([form indents alias-map]
    (transform form edit-all should-indent? #(indent-line % indents alias-map))))
 
+(defn split-maps
+  "Puts each hashmap pair on it's own line"
+  [form]
+  (transform form edit-all z/map?
```
weavejester, line 347: Are we searching for `z/map?`, or would a custom function like `map-value?` be a better target? If you write a search function that finds map values, then the logic for transforming becomes simpler in turn.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -341,6 +341,25 @@
...
+             (fn [zloc]
+               (let [split
+                     (loop [z (zip/children zloc) i 1 accumulate []]
+                       (let [nodes (if (= 0 (mod i 4))
+                                     ; 4th elements are the whitespace between pairs
```
weavejester, line 352: Can we guarantee the 4th element is the whitespace between pairs? What about a map like:

```clojure
{:x 1 ;; some comment
 :y 2}
```

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -926,7 +926,12 @@
         " bar "
         " )"]
        {:remove-surrounding-whitespace? false
-        :remove-trailing-whitespace? false})))
+        :remove-trailing-whitespace? false}))
+  (is (reformats-to?
+       ["{:one two :three four}"]
```
weavejester, line 931: This needs more tests to check edge conditions. What happens with comments, with commas, with `#_`, etc.

doby162:
> Thanks for the PR. This looks like a good feature to add, and the code is a good start, but has some issues that need to be resolved before it can be merged.

Thank you for the helpful feedback, I've made a few updates!
The tests cover a wide range of comment and comma related scenarios, which lead me down the path of inconsistencies between rewrite-clj and rewrite-cljs, but it's all working and tested against a few clojure source files in addition to the unit tests.

Take a look!

weavejester:
Thanks for the update. The extra tests are much appreciated.

In terms of the actual code, however, I think you're making this more difficult than it needs to be. Let's start by considering your top level transform:

```clojure
(transform form edit-all multiple-keys-per-line? split-keypairs))
```

You're targetting the maps, but the problem becomes simpler when you target the map values themselves. Consider instead:

```clojure
(transform form edit-all map-val-without-line-break? z/append-newline))
```

Now we just need to worry about how to identify the right zipper location:

```clojure
(defn map-val? [zloc]
  (and (z/map? (z/up zloc)) (odd? (index-of zloc))))

(defn map-val-without-line-break? [zloc]
  (when-let [r (zip/right zloc)]
    (and (map-val? zloc) (not (line-break? r)))))
```

I haven't tested this code, so there may be things I haven't considered, but this is the approach I'd take.

doby162:
Hey, I messed with this a bit last night. I got this map-value? based code to work with a couple tweaks, but since it appends newlines directly after map values, it results in awkward comma and comment displacement.

So, if we want to do this based on map elements instead of maps, I think we need a function preceded-by-newline? to check if an element has a newline between it and the nearest left non-whitespace token. Then we can just prepend the linebreaks to the keys and maintain the comma and comment support of the previous version of my patch.

Let me know if that sounds like a good plan, I might have some more time later this week.
Thanks!

weavejester:
Yep, that sounds fine. Doing it the other way around, checking keys for preceding linebreaks instead of values for proceding linebreaks, seems like a good solution.

doby162:
It's rewritten with element-level formatting instead of wholesale hashmap reformatting! Hope you like it-- let me know if you'd like any other tweaks.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -341,6 +341,28 @@
...
+      false
+      (line-break? previous)
+      true
+      (whitespace-or-comment? previous)
+      (recur (zip/left previous))
+      :else
+      false)))
+
+(defn map-key-without-line-break? [zloc]
+  (and (map-key? zloc) (zip/left zloc) (not (preceded-by-newline? zloc))))
```
weavejester, line 360: Why have `(zip/left zloc)` here? Isn't that covered by `(not (preceded-by-newline? zloc))`?

README.md
```diff
@@ -132,6 +132,10 @@ selectively enabled or disabled:
   true if cljfmt should collapse consecutive blank lines. This will
   convert `(foo)\n\n\n(bar)` to `(foo)\n\n(bar)`. Defaults to true.
 
+* `:split-keypairs-over-multiple-lines?` -
+  true if cljfmt should break hashmaps onto multiple lines.
```
weavejester, line 136: Lines in the README should be wrapped at 72 characters.

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -926,7 +926,72 @@
...
+  (is (reformats-to?
+       ["{:one two,"
+        " :three four}"]
+       ["{:one two,"
+        " :three four}"]
+       {:split-keypairs-over-multiple-lines? true
+        :remove-trailing-whitespace? false}))
+  (is (reformats-to?
+        ;this behavior is "wrong", but since rewrite-clj treats #_comments
+        ;as values, the way to resolve this is to make a PR against rewrite-clj
```
weavejester, line 978: We could fix this in cljfmt by ignoring `:uneval` forms.

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -926,7 +926,72 @@
...
+       ["{:one two #_comment"
+        ":three four}"]
+       ["{:one two"
+        " #_comment"
+        " :three"
+        " four}"]
+       {:split-keypairs-over-multiple-lines? true}))
+  (is (reformats-to?
+       ;this isn't a valid map so the output shouldn't be a valid map
```
weavejester, line 987: I don't think we need to test this.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -341,6 +341,28 @@
...
+
+(defn split-keypairs-over-multiple-lines
+  [form]
+  (transform form edit-all map-key-without-line-break? #(zip/insert-left % (n/newlines 1))))
```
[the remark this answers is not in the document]
doby162: This isn't possible presently but I have submitted a pull request to rectify the error: https://github.com/xsc/rewrite-clj/pull/81

cljfmt/src/cljfmt/core.cljc
```diff
@@ -341,6 +341,28 @@
   ([form indents alias-map]
    (transform form edit-all should-indent? #(indent-line % indents alias-map))))
 
+(defn map-key? [zloc]
+  (and (z/map? (z/up zloc)) (even? (index-of zloc)) (not (whitespace-or-comment? zloc))))
+
+(defn preceded-by-newline? [zloc]
```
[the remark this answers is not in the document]
doby162: I changed the name of the function to be more consistent!
Your suggested rewrite of the function doesn't seem to work as-is, and I'm not sure how to get `skip` to actually skip.
I left the function body unchanged for now.

cljfmt/test/cljfmt/core_test.cljc
doby162: Done, I added to index-of to allow it to ignore uneval forms based on the definition in start-element

doby162:
Hey, just leaving a comment because I've not sure if my commit and individual issue commits triggered GitHub to consider this to need a new review! Let me know if you'd like any other tweaks!

cljfmt/src/cljfmt/core.cljc
```diff
@@ -179,8 +179,13 @@
 (defn- coll-indent [zloc]
   (-> zloc zip/leftmost margin))
 
+(defn uneval-node? [zloc]
```
weavejester, line 182: This should be `uneval?` to match other similar functions, it should be private, and it should use the tag to determine the node type rather than converting it to a string:

```clojure
(defn- uneval? [zloc]
  (= (z/tag zloc) :uneval))
```

cljfmt/src/cljfmt/core.cljc
```diff
@@ -179,8 +179,13 @@
...
+  (when (z/string zloc)
+    (str/starts-with? (z/string zloc) (:uneval start-element))))
+
 (defn- index-of [zloc]
   (->> (iterate z/left zloc)
+       (filter #(not (uneval-node? %)))
```
weavejester, line 188: A better way of writing this is:

```clojure
(remove uneval?)
```

cljfmt/src/cljfmt/core.cljc
```diff
@@ -341,6 +346,35 @@
   ([form indents alias-map]
    (transform form edit-all should-indent? #(indent-line % indents alias-map))))
 
+(defn map-key? [zloc]
+  (and
```
weavejester, line 350: Standard Clojure formatting should be used. So:

```clojure
(and (z/map? (z/up zloc))
     (even? (index-of zloc))
     (not? (uneval? zloc))
     (not (whitespace-or-comment? zloc)))

File: cljfmt/src/cljfmt/core.cljc [hunk 16]
```diff
@@ -341,6 +346,35 @@
...
+    (z/map? (z/up zloc))
+    (even? (index-of zloc))
+    (not (uneval-node? zloc))
+    (not (whitespace-or-comment? zloc))))
+
+(defn preceded-by-line-break? [zloc]
```
weavejester, line 356: This function should be private.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -341,6 +346,35 @@
...
+  (loop [previous (zip/left zloc)]
```
weavejester, line 357: This could be written more concisely as:

```clojure
  (loop [previous (zip/left zloc)]
    (if (whitespace-or-comment? previous)
      (recur (zip/left previous))
      (and previous (line-break? previous))))
```

cljfmt/src/cljfmt/core.cljc
```diff
@@ -341,6 +346,35 @@
...
+      (nil? previous)
+      false
+      (line-break? previous)
+      true
+      (whitespace-or-comment? previous)
+      (recur (zip/left previous))
+      :else
+      false)))
+
+(defn map-key-without-line-break? [zloc]
```
weavejester, line 368: This function should be private.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -341,6 +346,35 @@
...
+  (and (map-key? zloc) (not (preceded-by-line-break? zloc))))
+
+(defn split-keypairs-over-multiple-lines [form]
+  (transform
```
weavejester, line 372: This should use standard Clojure indentation, and could be more concisely written:

```clojure
(defn- insert-newline-left [zloc]
  (zip/insert-left zloc (n/newlines 1)))

(defn split-keypairs-over-multiple-lines [form]
  (transform form edit-all map-key-without-line-break? insert-newline-left))
```

README.md
weavejester: Wrapping is still incorrect. It should look like:

```
  true if cljfmt should break hashmaps onto multiple lines. This will
  convert `{:a 1 :b 2}` to `{:a 1\n:b 2}`. Defaults to false.
```

cljfmt/src/cljfmt/core.cljc
weavejester: Also `map-key?` should be private.

doby162:
Took care of this feedback! The `preceded-by-line-break?` function rewrite provided didn't quite work but I was able to tighten it up a little bit nonetheless!

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -926,7 +926,59 @@
...
+        " :three four}"]
+       ["{:one two"
+        " :three four}"]
+       {:split-keypairs-over-multiple-lines? true}))
+  (is (reformats-to?
+       ["{:one two"
+        ";comment"
+        ":three four}"]
+       ["{:one two"
+         ";comment"
```
weavejester, line 946: The indentation is off here, but other than that, I think the commits are good to be squashed down, ready for merge.

doby162:
Fixed the indentation!

weavejester:
Thanks! Can you squash down your commits?

doby162:
Sorry, I thought you meant you were going to to a squash merge! I have rebased the commits into one, ready to go.

Thanks!

weavejester:
Merged. Thanks!

PR #196, opened 2020-05-04
Adding :more-newlines? option

ertugrulcetin:
Cljfmt's default behaviour is putting 2 newlines between forms (defns), it's good to have 3 newlines, makes it easy to read.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -430,7 +430,15 @@
 (defn find-line-separator [s]
   (or (re-find #"\r?\n" s) default-line-separator))
 
-(defn wrap-normalize-newlines [f]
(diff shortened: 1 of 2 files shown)


weavejester:
> Cljfmt's default behaviour is putting 2 newlines between forms (defns)

It doesn't actually do that at present. It just has an option to ensure that multiple newlines don't exceed two.

ertugrulcetin:
@weavejester does this PR make sense to you or not needed feature?

weavejester:
I don't think this PR makes sense in this form. If you want more spacing between forms, you can disable the option to remove consecutive blank lines. At present, cljfmt doesn't have any option to enforce spacing between top-level forms.

ertugrulcetin:
Alright, makes sense, thanks! :)

Issue #190, opened 2020-03-03
Can't use all indent configuration options via tools.deps


mynomoto:
First, thanks for supporting tools.deps as an option to run cljfmt.

I was trying to migrate cljfmt from using the lein plugin to tools.deps on one project but since the indents file is parsed as edn it doesn't support regexes.
```
[Error while parsing option "--indents indentation.edn": java.lang.RuntimeException: No dispatch macro for: "]
```
Would a change to allow regexes be welcome?

hoxu:
I guess it's not possible to configure this example using edn currently:

    {:indents ^:replace {#".*" [[:inner 0]]}

mynomoto:
> I guess it's not possible to configure this example using edn currently:
> 
> ```
> {:indents ^:replace {#".*" [[:inner 0]]}
> ```

Yeah, sorry for not providing a example when I opened the issue.

hoxu:
@weavejester @mynomoto: any workarounds or ideas how to fix this?

mynomoto:
So far I'm still using lein to avoid the issue. I think you can build your own function to read a clj file, I just wanted feature parity out of the box.

weavejester:
Sorry for the late response. Perhaps the best solution to this would be to allow `--indents` to reference a `.clj` file as well as a `.edn` file, and use the appropriate `read-string` depending on the file extension.

mynomoto:
> Sorry for the late response. Perhaps the best solution to this would be to allow --indents to reference a .clj file as well as a .edn file, and use the appropriate read-string depending on the file extension.

This would be a nice solution. Would a pr be welcome?

weavejester:
> This would be a nice solution. Would a pr be welcome?

Yep.

lread:
@weavejester can this be closed?  Looks like a solution was merged?

weavejester:
Yep, this looks fixed.

PR #202, opened 2020-08-18
Use read-string if .clj file when parsing cli opts

frankitox:
I'm trying to fix the problem described in #190. It's just a couple of lines of code. Let me know what you think.


cljfmt/src/cljfmt/diff.clj
```diff
@@ -16,6 +16,9 @@
        ^java.io.File (apply io/file)
        .getCanonicalPath))
 
+(defn filename-ext [filename]
```
weavejester, line 19: Could you move this function to main.clj (and make it private)? It doesn't fit in diff.clj.

cljfmt/src/cljfmt/main.clj
```diff
@@ -136,8 +136,12 @@
              (warn "Failed to format file:" (project-path options f))
              (print-stack-trace e))))))))
 
-(def ^:private cli-file-reader
-  (comp edn/read-string slurp diff/to-absolute-path))
+(defn- cli-file-reader [opt-value]
```
weavejester, line 139: Instead of `opt-value`, prefer a more descriptive name like `filepath`.

cljfmt/src/cljfmt/main.clj
```diff
@@ -136,8 +136,12 @@
...
+  (let [filename (diff/to-absolute-path opt-value)
```
weavejester, line 140: What's the purpose of converting the filename to an absolute path?

weavejester (changes requested):

frankitox:
Hi @weavejester, thanks for taking the time to review this! These are the changes I just made:

- [x] Moved `relative-path` to `cljfmt.main` and make it private (let me know if you prefer other name for `relative-path`).
- [x] Renamed `opt-value` to `filepath`.

Note that `cli-file-reader` is also used by `--alias-map`. Which means you can also pass `.edn` files to that parameter. Is that ok with you?

weavejester:
It seems fine to be consistent and have all options support both .edn and .clj file types.

frankitox:
Gotcha. Give me a moment because I force pushed instead of creating a new commit, so you can actually see what I changed.

frankitox:
There we go, you can check the [last changes](https://github.com/weavejester/cljfmt/pull/202/commits/ffc2dcc11eec599a5b6abb255b5e57330722c17f) now. Thanks!

weavejester:
Thanks! I also had a question on why you're using `diff/to-absolute-path` in `cli-file-reader`. It doesn't look necessary to me.

cljfmt/src/cljfmt/main.clj
frankitox: Well, I just blindly copied `cli-file-reader`'s source code. I made a slight [change](https://github.com/weavejester/cljfmt/pull/202/commits/ffc2dcc11eec599a5b6abb255b5e57330722c17f#diff-a57e762da42583224b62656b3ae05addR144-R145), so `filename-ext` receives the raw string, but let me know if you meant something else.

weavejester: Ah, I see. I'm not sure why the `diff/to-absolute-path` is there to begin with. I think we can safely remove it.

frankitox: Great! Yes, I can't see a reason for it either. I just [removed it](https://github.com/weavejester/cljfmt/pull/202/commits/334ef64029fb132249b89f7f06af3d8aa0d76a8f).

PR #203, opened 2020-08-21
Add :remove-extra-whitespace? option

mpenet:
Hi James, 

This pr adds a `:remove-extra-whitespace?` (defaulting to false). 

This can be handy to turn a project that used justified form alignement into the classical form.

It will go through all whitespace nodes and replace the ones with more than 1 char with a single char whitespace node. 
I set the default to false has not to be too surprising since now cljfmt will deal both styles (justified or not) by not altering the result. It will leave comments untouched. 

One quirky thing is that it must/should be run with `:indentation? true` only, otherwise the source gets all "compressed", so there's an argument to make it an option for indentation? or maybe force indentation if that formatter is on, not sure about that one. 

```text
[a    b c] -> [a b c]
{:a    :b} -> {:a :b} 
```

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -386,6 +386,15 @@
 (defn remove-trailing-whitespace [form]
   (transform form edit-all trailing-whitespace? zip/remove))
 
+(defn- replace-with-one-space [zloc]
+  (zip/replace zloc (whitespace 1)))
+
+(defn- non-indenting-whitespace? [zloc]
+  (and (whitespace? zloc) (not (indentation? zloc))))
+
+(defn remove-multiple-non-indenting-spaces [form]
+  (transform form edit-all non-indenting-whitespace? replace-with-one-space))
+
 (defn reformat-form
   ([form]
    (reformat-form form {}))
@@ -399,6 +408,8 @@
          remove-surrounding-whitespace)
        (cond-> (:insert-missing-whitespace? opts true)
          insert-missing-whitespace)
+       (cond-> (:remove-multiple-non-indenting-spaces? opts false)
+         remove-multiple-non-indenting-spaces)
        (cond-> (:indentation? opts true)
          (reindent (:indents opts default-indents)
                    (:alias-map opts {})))
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -627,6 +627,44 @@
          ["#:clj {:a :b"
           "       :c :d}"]))))
 
+(deftest test-remove-multiple-non-indenting-spaces
+  (let [opts {:remove-multiple-non-indenting-spaces? true}]
+    (is (reformats-to? ["[]"] ["[]"] opts))
+    (is (reformats-to? ["[   ]"] ["[]"] opts))
+    (is (reformats-to? ["{   }"] ["{}"] opts))
+    (is (reformats-to? ["#{ }"] ["#{}"] opts))
+    (is (reformats-to? ["[a     b]"] ["[a b]"]  opts))
+    (is (reformats-to? ["{a     b}"] ["{a b}"] opts))
+    (is (reformats-to? ["{a,     b}"] ["{a, b}"] opts))
+    (is (reformats-to? ["#{a     b}"] ["#{a b}"] opts))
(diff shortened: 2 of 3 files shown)


mpenet:
> Can't you use this option to fix the spaces that are not part of the indentation, and use the indentation fixer to handle the indentation spaces?

Done as you suggested with latest commit.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -386,6 +386,27 @@
...
+
+(defn- prev-indenting?
+  "Returns true if all prev elements in line (if any) are whitespaces"
+  [zloc]
+  (or (newline? zloc)
+      (comment? zloc)
+      (and (whitespace? zloc)
+           (recur (zip/prev zloc)))))
+
+(defn- non-indenting-spaces? [zloc]
```
weavejester, line 403: This function could be written more succinctly as:

```clojure
(defn- non-indenting-space? [zloc]
  (and (whitespace? zloc) (not (indentation? zloc))))
```

Also we probably want to call it `non-indenting-whitespace?` come to think of it.

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -627,6 +627,34 @@
...
+    (is (reformats-to? ["#{    }"] ["#{}"] opts))
+    (is (reformats-to? ["[a     b   c]"] ["[a b c]"] opts))
+    (is (reformats-to? ["#{a     b     }"] ["#{a b}"] opts))
+    (is (reformats-to? ["(cond truc"
+                        "      ;; foo"
+                        "      (bar? a    b) :baz)"]
+                       ["(cond truc"
+                        "      ;; foo"
+                        "      (bar? a b) :baz)"] opts)
+        "ensure comments are respected")
```
weavejester, line 648: Remove "ensure " from description. "Testing comments are respected" works better than "Testing ensure comments are respected."

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -627,6 +627,34 @@
...
+    (is (reformats-to? ["(cond truc"
+                        "             ;; foo"
+                        "   (bar? a    b) :baz)"]
+                       ["(cond truc"
+                        "             ;; foo"
+                        "   (bar? a b) :baz)"]
+                       (assoc opts :indentation? false))
+        "ensure custom indentation is respected")))
```
weavejester, line 656: There should also be a test to ensure non-comment newlines work.

weavejester (changes requested):

mpenet:
amended as per comments

weavejester:
Thanks! How is this option expected to work around forms with commas? For example: `{:a 1, :b 2}`.

mpenet:
> Thanks! How is this option expected to work around forms with commas? For example: `{:a 1, :b 2}`.

I think it should retain them. I added a test case to reflect this.

mpenet:
@weavejester let me know if you need any other changes to the PR. 

Something I was not sure about is making this a default or not, I guess it could be considered quite disruptive for people using "justified" alignment.

weavejester:
It shouldn't be the default, as the default should just just fix code that doesn't adhere to the Clojure style guide.

Can you change the commit message to:

```
Add :remove-multiple-non-indenting-spaces option

This option replaces all non-indenting horizontal whitespace with a
single space. Commas are preserved. This option is off by default.
```

cljfmt/src/cljfmt/core.cljc
```diff
@@ -386,6 +386,16 @@
 (defn remove-trailing-whitespace [form]
   (transform form edit-all trailing-whitespace? zip/remove))
 
+(defn- replace-with-one-space [zloc]
+  (zip/replace zloc (whitespace 1)))
+
+(defn- non-indenting-whitespace? [zloc]
+  (and (whitespace? zloc)
```
weavejester, line 393: Could you also put this on one line, so:

```clojure
(defn- non-indenting-whitespace? [zloc]
  (and (whitespace? zloc) (not (indentation? zloc))))
```

mpenet:
Done and squashed!

weavejester:
Merged. Thanks for the PR.

mpenet:
> Merged. Thanks for the PR.

Thanks!

Issue #176, opened 2019-09-23
Format ns-qualified maps?


vemv:
Hi there,

cljfmt doesn't appear to format ns-qualified maps:

```
cljfmt.test-util.common=> (reformat-string "#:foo.bar  {:owner-type 1}" {})
"#:foo.bar  {:owner-type 1}"
cljfmt.test-util.common=> (reformat-string "#:foo.bar {:owner-type 1}" {})
"#:foo.bar {:owner-type 1}"
cljfmt.test-util.common=> (reformat-string "#:foo.bar{:owner-type 1}" {})
"#:foo.bar{:owner-type 1}"
```

...this has the specific impact (at least as my coworker reports) that Cursive's Parinfer can stop working properly if there's any whitepsace between the qualifer and the map.

Is this something you have tried working on at some point? (Asking because it might be hard to fix)

Thanks - V

weavejester:
If someone wants to send me a PR to add a feature then I'll certainly review and (eventually!) merge it in. But I have to many other projects to devote much time to cljfmt myself.

lread:
@vemv, I think I might have started on this a long while ago [but abandoned my effort](https://github.com/weavejester/cljfmt/pull/158). I can take up the cause again if/after we [upgrade to rewrite-clj v1](https://github.com/weavejester/cljfmt/issues/211).

PR #213, opened 2021-03-17
Upgrade to rewrite-clj v1

### LINKED ISSUE #211: Consider upgrading to rewrite-clj v1
I've recently released [rewrite-clj v1](https://github.com/clj-commons/rewrite-clj/blob/main/CHANGELOG.adoc#rewrite-clj-v1). 

Even though it is labelled, alpha, it is feeling solid.
I am already [successfully running cljfmt's tests against rewrite-clj v1](https://github.com/clj-commons/rewrite-clj/blob/main/doc/02-developer-guide.adoc#libs-test).

Upgrading to rewrite-clj v1 should fix many cljfmt issues that were previously waiting on fixes in rewrite-clj v0 and rewrite-cljs.

Potentially problematic? (lemme know!):
- rewrite-clj v1 min Clojure version is 1.9
- rewrite-clj v1 ClojureScript performance could be less than that of rewrite-cljs. I dropped rewrite-cljs specific tunings in favour of a single code base. I do plan to address any real performance issues.

I am happy to:
- Compare cljfmt test suite execution time with rewrite-cljs vs rewrite-clj v1 to get some idea of potential degradations.
- Create a rewrite-clj v1 upgrade PR which will include a review current open cljfmt issues, and when appropriate, the addition of tests to verify that they are fixed by rewrite-clj v1

lread:
For your review, an upgrade to rewrite-clj v1.

I've tried to tell the story through the commits.

Happy to do separate PRs for the revert and fixes, if that makes more sense to you.

Closes #211

### DIFF
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -1049,7 +1049,11 @@
       "regular expression with embedded newline")
   (is (reformats-to? 
        ["##Inf ##-Inf ##NaN"] 
-       ["##Inf ##-Inf ##NaN"])))
+       ["##Inf ##-Inf ##NaN"]))
+  (is (reformats-to?
+       [":&::before"]
+       [":&::before"])
+      "garden library style selector"))
 
 (deftest test-namespaced-keywords
   (is (reformats-to? ["(ns myns.core)"
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -1068,6 +1068,20 @@
   (is (reformats-to? ["(println (not (= ::result/foo 2)))"]
                      ["(println (not (= ::result/foo 2)))"])))
 
+(deftest test-namespaced-maps
+  (is (reformats-to?
+       ["#:my-prefix    {:a 1}"]
+       ["#:my-prefix    {:a 1}"]))
+  (is (reformats-to?
+       ["#:my-prefix{:a 1}"]
+       ["#:my-prefix{:a 1}"]))
+  (is (reformats-to?
+       ["#::{:b 2}"]
+       ["#::{:b 2}"]))
+  (is (reformats-to?
+       ["#::my-alias{:c 3}"]
+       ["#::my-alias{:c 3}"])))
+
 (deftest test-normalize-newlines
   (is (= (normalize-newlines "foo\nbar\nbaz") "foo\nbar\nbaz"))
   (is (= (normalize-newlines "foo\r\nbar\r\nbaz") "foo\nbar\nbaz"))
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -1046,7 +1046,10 @@
        ["#\"(?i)foo\""]))
   (is (= "#\"a\nb\""
          (reformat-string "#\"a\nb\""))
-      "regular expression with embedded newline"))
+      "regular expression with embedded newline")
+  (is (reformats-to? 
+       ["##Inf ##-Inf ##NaN"] 
+       ["##Inf ##-Inf ##NaN"])))
 
 (deftest test-namespaced-keywords
   (is (reformats-to? ["(ns myns.core)"
@@ -1060,7 +1063,6 @@
                       "               (foo))}"]))
   (is (reformats-to? ["(println (not (= ::result/foo 2)))"]
                      ["(println (not (= ::result/foo 2)))"])))
-
 
 (deftest test-normalize-newlines
   (is (= (normalize-newlines "foo\nbar\nbaz") "foo\nbar\nbaz"))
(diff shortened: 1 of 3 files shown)


weavejester:
Thanks for the PR! The separation of commits is good, but the wording could be adjusted, e.g. I didn't understand that "verify" meant "add test". Let's come to that last of all though.

Overall the code changes look fine, but why did you replace functions like `zip/next` with `z/next*`?

lread:
Thanks for the review!  

> but the wording could be adjusted, e.g. I didn't understand that "verify" meant "add test". Let's come to that last of all though.

Yeah, I was trying to suggest that I wasn't adding the fix in the commit, just the test, but perhaps was more wordy than I needed to be. Perhaps a violation of [rule #7?](https://chris.beams.io/posts/git-commit/#why-not-how). I'm always ready for, and appreciate, lessons in succinctness when I work on cljfmt and am happy to refine.

> but why did you replace functions like `zip/next` with `z/next*`?

The `*` variants have been in the rewrite-clj zip API since v0, but they were not in rewrite-cljs, nor were they described in rewrite-clj v0 docs. Technically, they provide raw access to the underlying zipper.

To me they've become synonymous with "don't do whitespace skipping/processing" and the fact that they even use an underlying zipper or there is an underlying zipper is, in my mind, just an implementation detail.

I think they were added when the position tracking zipper support was added to rewrite-clj v0 (a feature that cljfmt does not use). Position tracking uses an internal custom zipper rather than `clojure.zip`.

That all said, because cljfmt is not using rewrite-clj's position tracking, `clojure.zip/next`, for example, can still be used instead of `rewrite-clj.zip/next*`.

weavejester:
Okay, that's sounds fine, then.

With regard to the commits, the first one should be something like:

```
Update rewrite-clj to 1.0 alpha

Minimum Clojure version updated to 1.9.
```

This commit should still include rewrite-cljs, as we can include that as part of the next commit

```
Replace rewrite-cljs with rewrite-clj

Version 1.0 of rewrite-clj supports ClojureScript.
```

This commit should remove the rewrite-cljs dependency at the same time as removing the requires.

```
Revert fix for ClojureScript comma bug 

No longer necessary under rewrite-clj 1.0.
```

The details aren't neccessary. Just what the commit does (revert the fix), and an explanation why (it's no longer needed).

```
Add tests for namespaced keywords in ClojureScript

Closes #106.
```

"Add tests" is a less ambiguous term than "Verify". Similarly:

```
Add tests for ##Inf ##-Inf and ##NaN

Closes #171.
```

lread:
Thanks @weavejester, very succinct, indeed!  The only thing I will change is merging your first 2 commits into 1.  Rewrite-clj v1 replaces both rewrite-clj v0 and rewrite-cljs. If I leave rewrite-cljs in while also depending on rewrite-clj v1, we'd have, only very temporarily, mind you, 2 ClojureScript `rewrite-clj` namespaces on the classpath after the 1st commit.

weavejester:
Thanks. In which case, let's change the explanation a little to give a little more context.

```
Update rewrite-clj to 1.0 alpha

Removes rewrite-cljs dependency, as version 1.0 of rewrite-clj supports
ClojureScript. Also updates minimum Clojure version to 1.9.
```

lread:
Thanks, done!

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -1045,7 +1046,23 @@
...
+      "regular expression with embedded newline")
+  (is (reformats-to? 
+       ["##Inf ##-Inf ##NaN"] 
+       ["##Inf ##-Inf ##NaN"])))
+
+(deftest test-namespaced-keywords
+  (is (reformats-to? ["(ns myns.core)"
+                      ""
+                      ""
+                      "{::foo-bar (do (map)"
```
vemv, line 1058: I wonder if it's worth adding test coverage for the following related constructs...?

* current-ns qualified map: `#::{:b 2}`
* ns-alias qualified map: `#::alias{:c 3}`

They might as well deserve a separate PR ([related issue](https://github.com/weavejester/cljfmt/issues/176)), but my concern here would be that upgrading rewrite-clj could break something and was previously not broken.

lread: @vemv I think this is a really good feedback, thanks.
I will add some more tests in as part of this PR today.
From my notes, [rewrite-cljs and rewrite-clj v0 could not even parse current namespace autoresolve namespaced maps ](https://github.com/clj-commons/rewrite-clj/blob/main/CHANGELOG.adoc#v10572-alpha-1)(`::{:b 2}`), so I'll try the tests first against cljfmt master.

lread: Tests added

lread:
@weavejester, I have re-reviewed rewrite-clj v1 changes and added additional tests to cljfmt (namespaced maps and garden style selectors).

Also of interest:
- rewrite-clj v1 normalizes newlines to `\n`.
This, as far I  understand it, is the intent of the clojure tools reader but something [it currently stumbles on](https://clojure.atlassian.net/browse/TRDR-65).
Because it was causing failures in parsing, [I compensated for tools reader in rewrite-clj v1](https://github.com/clj-commons/rewrite-clj/commit/f78856edbb9d26a9481eeaf33d071bc3d8ad8868).
I've not looked into how cljfmt handles OS specific line endings, but will have a look if this is raising any concerns for you.
- rewrite-clj v1 [now preserves a newline, if present, at end of file](https://github.com/clj-commons/rewrite-clj/issues/121).
I don't expect this is a concern, but thought I should raise it for completeness.

weavejester:
This all looks fine, as long as you've finished updating it.

lread:
Thanks @weavejester, rewrite-clj v1 re-upgrade commit squashed, looks all good to go from my side.

PR #219, opened 2021-04-01
Add report-unsupported-native-image-at-runtime flag

### LINKED ISSUE #217: #202 breaks native-image build
Following the request in #210 I came to review my recent changes and found that native image building is broken.
Analysis:
Checked out commit `7571de0`, built a native-image with `lein do clean, native-image`, native image built successfully
Checked out commit `98a7c53`, built a native-image with `lein do clean, native-image`, native image build failed
Possible causes:
`read-string` calls unsupported method `java.lang.ClassLoader.defineClass(String, byte[], int, int)`

@borkdude what do you use in babashka for reading strings? From looking at the code base I see only `edn/read-string` is used. It is sufficient for this use case?

bsless:
read-string breaks native-image building due to an unsupported reachable
method. By delaying the reporting of this method to run-time, it turns
out the compiled native-image still works.

Closes #217.

### DIFF
--- a/cljfmt/project.clj
+++ b/cljfmt/project.clj
@@ -29,6 +29,7 @@
           "-J-Dclojure.compiler.direct-linking=true"
           "-H:ReflectionConfigurationFiles=reflection.json"
           "--initialize-at-build-time"
+          "--report-unsupported-elements-at-runtime"
           "-H:Log=registerResource:"
           "--verbose"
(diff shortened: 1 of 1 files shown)


weavejester:
I think the second paragraph of the commit message can be cut out, as that's easily found via a search.

"Closes #217" should end with a period, ".".

Issue #49, opened 2015-11-29
Support var indentation metadata


bbatsov:
We've come up with some indentation spec for CIDER's dynamic indentation functionality, we consider fairly flexible and not CIDER-specific at all. The details are [here](https://github.com/clojure-emacs/cider/blob/master/doc/indent_spec.md#indent-specification).

It'd be nice if `cljfmt` (and other tools in general) respected indentation settings provided via such metadata. //cc @malabarba

Malabarba:
Indeed. Would be nice if this could be supported by cljfmt too. 
If you decide to try, I'd be happy to help as well.

weavejester:
This seems like a good idea. I think the indentation spec could potentially replace cljfmt's own, less well thought out syntax. I don't think there's anything I'm doing that isn't covered by the linked spec.

mitchelkuijpers:
I would like to take a stab at this @weavejester Would you still consider this? @Malabarba I think I'll take you up on the help ^^

arrdem:
@mitchelkuijpers awesome! Thanks for taking this up, I glanced at it a few weeks ago and didn't get anywhere. As you work on this, I'd keep a critical eye on both indentation spec formalisms. I'm concerned that for instance @weavejester likes to write

```clojure
(ns
  (:require
   []
   ...))
```

Neither indentation specifier provides a way to require linebreaks.

Also as far as I'm able to tell, neither indentation specifier is able to support repetitions of term(s). I haven't been able to find the particular thread with some quick googling, but a fellow came along who was using the following spec for cond: `[2 4 2 4 2 4 2 4 ...]` because he wanted result expressions to be on the next line and indented more deeply than the condition expressions.

I'd be happy to help if I can.

weavejester:
> I'm concerned that for instance @weavejester likes to write
> ```clojure
> (ns
>   (:require
>    []
>    ...))
```

Actually I use:

```clojure
(ns foo.bar
  (:require [foo.baz :as baz]
            [foo.quz :as quz]
            ...))
```

The lines in `cljfmt.core` weren't written by me. I think I allowed that style because otherwise the lines would have been too long.

> Neither indentation specifier provides a way to require linebreaks.

Well, because they're indentation specifiers, not linebreak specifiers. Automatically adding linebreaks is another problem.

arrdem:
[...]

PR #230, opened 2021-05-19
Parse style/indent metadata [WIP]

camsaul:
Implements #49

@weavejester, let me know if this approach works for you and I can add some tests and documentation and figure out how to translate more complicated `:style/indent` specs.

The basic idea here is to load the namespace associated with a file when we go to reformat/check it and then look in all the vars defined in that namespace and in its `ns-aliases` namespaces for `:style/indent` metadata, then take the specs and merge them into the `indents` map that gets passed to `cljfmt.main/reformat-string`. If it can't load the namespace for one reason or another (e.g. because the files aren't on the classpath) it will just catch the exception and reformat the file the same way it did before this PR.

I didn't want to duplicate all this code between both the `check` and `fix` code paths so I abstracted out that stuff into a new `cljfmt.files` namespace. Both `check` and `fix` transduce a `cljfmt.files/reducible-files` Eduction that includes all the relevant info as well as some other stuff that previously had to be recalculated on the fly. So overall I think this simplifies the code in `cljfmt.main` a bit.

I also made some tweaks to `check` and `fix` so they don't call `System/exit` if they fail. Instead, they throw an Exception, which `-main` catches and prints, then exits with a nonzero status. I made this change to make development easier (I was testing stuff out in the REPL as I worked on this) and so `cljfmt` is useful interactively -- supporting calling `cljfmt/check` inside a running REPL in a larger project seems like a nice thing to support, similar to how I can run `clojure.test` tests or Eastwood programmatically.

I pushed a working `deps.edn` demo to https://github.com/camsaul/cljfmt that you can try with

```bash
clojure -Sdeps \
  '{:deps {camsaul/cljfmt {:git/url "https://github.com/camsaul/cljfmt", :sha "ec3884db9b0961c59fd74972c1ca4594f0ab477f"}}}' \
  -M -m cljfmt.main [check|fix]
```

One last note: I haven't coded up support for every single indentation spec variation supported by CIDER `:style/indent` yet. So far it supports numeric indent specs and `:defn`, but not more complicated stuff like `[1 :defn]` or the like -- I haven't really dug in to how to translate all the specs at this point. If it doesn't understand a spec, it currently just prints a warning and moves on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,8 @@ It is possible to execute `cljfmt` using the
 [Clojure CLI]: https://clojure.org/guides/deps_and_cli
 
 ```bash
-clojure -Sdeps '{:deps {cljfmt {:mvn/version "0.6.4"}}}' \
-  -m cljfmt.main [check|fix]
+clojure -Sdeps '{:deps {cljfmt/cljfmt {:mvn/version "0.7.0"}}}' \
+  -M -m cljfmt.main [check|fix]
 ```
 
 Customizing the rules is possible passing `edn` formated files,
@@ -219,7 +219,7 @@ Indentation types are:
 * `:inner` -
   two character indentation applied to form arguments at a depth
   relative to a form symbol
-  
+
 * `:block` -
   first argument aligned indentation applied to form arguments at form
   depth 0 for a symbol
@@ -234,9 +234,9 @@ Form depth is the nested depth of any element within the form.
 A contrived example will help to explain depth:
 
 ```clojure
-(foo 
+(foo
  bar
- (baz 
+ (baz
   (qux plugh)
   corge)
  (grault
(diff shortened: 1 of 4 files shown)


weavejester:
Thanks for the PR, but I'm afraid this approach isn't viable.

In order to support CIDER-style indentation metadata, we first need to fully support the syntax through the existing indentation configuration. Most likely this will require rewriting the indentation functions in `cljfmt.core` to use CIDER's syntax.

Once we have that up and working and fully tested, then we can approach the problem of pulling out metadata. Currently cljfmt does not evaluate the code it formats, and I'd like to keep it that way. Therefore we'd need a function that, given a symbol name and a classpath, returns the corresponding metadata without evaluating the namespace. Obviously this won't be foolproof - as macros exist and vars are mutable - but it would catch the vast majority of use-cases.

camsaul:
Ok, makes sense

Issue #232, opened 2021-05-21
Any interest in a clj-kondo config for cljfmt developers?


lread:
Hi @weavejester, I don't know if you use the [clj-kondo](https://github.com/clj-kondo/clj-kondo) linter.
I do and find it quite helpful.
I include it in my build processes and fail builds that have lint errors.

If you are interested, I am happy to do the work.

weavejester:
I had a quick check over it, and it looks like a very useful tool to be used manually, but in the case of cljfmt in particular, it has too many false positives to be included in the build chain.

lread:
Thanks for taking a peek @weavejester. 

Clj-kondo is highly configurable, so we could configure it to enforce whatever opinions are important to you.

Please ignore the remainder of this comment if you have already made up your mind and decided clj-kondo is not a good fit for clfjmt. Personally, I find the benefit of clj-kondo outweighs the cost of maintaining its config, but that's just me and might not be you.

Here's a sample clj-kondo usage:

Initial setup (from cljfmt root dir):

```
cd cljfmt
mkdir .clj-kondo
clj-kondo --lint "$(lein classpath)" --dependencies --parallel --copy-configs
```

And a run:
```
> clj-kondo --lint src test
src/cljfmt/core.cljc:18:8: warning: Unused private var cljfmt.core/find-all
src/cljfmt/core.cljc:74:48: warning: #'rewrite-clj.zip/append-space is deprecated since 0.5.0
src/cljfmt/core.cljc:207:8: warning: Unused private var cljfmt.core/top-level-form
src/cljfmt/core.cljc:227:3: warning: Missing else branch.
src/cljfmt/core.cljc:233:3: warning: Missing else branch.
src/cljfmt/core.cljc:262:5: warning: Missing else branch.
src/cljfmt/core.cljc:269:21: warning: Missing else branch.
src/cljfmt/core.cljc:282:3: warning: Missing else branch.
src/cljfmt/core.cljc:296:8: warning: unused binding sym
src/cljfmt/core.cljc:296:12: warning: unused binding alias-map
src/cljfmt/core.cljc:296:30: warning: unused binding args
src/cljfmt/diff.clj:2:31: warning: Unused import Delta$TYPE
src/cljfmt/main.clj:202:8: warning: Unused private var cljfmt.main/command-name
test/cljfmt/core_test.cljc:7:14: warning: namespace cljfmt.test-util.common is required but never used
[...]

PR #233, opened 2021-05-23
Refactor test utilities

lread:
Obviate the need for a compensating namespace require in core-test
for ClojureScript by switching from a function to a macro for common
code supporting reformats-to? assert-expr.

Rename test_util/cljs.cljc to test_util/cljs.clj to avoid unnecessary
complexity.

### DIFF
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -4,7 +4,6 @@
             [cljfmt.core :refer [reformat-string default-line-separator
                                  normalize-newlines find-line-separator
                                  replace-newlines wrap-normalize-newlines]]
-            [cljfmt.test-util.common :as common]
             #?(:clj [cljfmt.test-util.clojure]))
   #?(:cljs (:require-macros [cljfmt.test-util.cljs])))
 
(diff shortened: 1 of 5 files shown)


weavejester:
I'm not sure I understand the purpose of this PR. What benefit does this have?

lread:
Hi @weavejester, it stemmed from our discussions in #232.
Just a code cleanup, no behaviour changes.

No longer any need for that confusing what seemed to be unused require from core-test.
And there wasn't a need for cljs.cljc to be cljc so now it is clj.

weavejester:
I don't think "unused" requires are confusing. If you want to extend a third-party multi-method or protocol, then it's not uncommon to have a namespace dedicated to that. I don't think removing a require is worth changing a function into a macro.

lread:
Oh ok. I felt that this was an improvement to my original stab at the assert-expr years ago.

But, this is subjective, it is your repo, and I do respect your decision to disagree and close this PR.

weavejester:
I think I'd weight having less code in macros as being more important than avoiding namespaces without vars. At least in my view, the former adds more complexity than the latter.

lread:
Certainly valid.

For me assert-expr is a bit of black art. 
I do not find it terribly straightforward to to write an assert-expr that works under both Clojure and ClojureScript.
My attempt here was to appease the way ClojureScript seems to behave. 
I was proud to have appeased the beast, but I do see how it could be viewed as replacing one work-around with an inferior work-around.

weavejester:
I agree there's no good answers here, and I appreciate your attempt. `clojure.test` is one of the weaker parts of the standard library, I feel.

lread:
Thanks for your review and patience!

Issue #226, opened 2021-04-24
reformat-form performance issues


ericdallo:
This is a separate issue regarding the performance issues mentioned on #212.

Context:
On [clojure-lsp](https://clojure-lsp.github.io/clojure-lsp/) we use cljfmt to 2 LSP features: 
- format the whole buffer (using `reformat-form` with the buffer as string)
- format a range that can be triggered in multiple ways, like the user selecting a range and triggering format or the most common, using any parenthesis manipulation will trigger a reformat on that range (using `reformat-form`).

The issue seems that reformat-form can be very slow on medium/big forms/buffers affecting the parenthesis manipulation performance for users like [here](https://github.com/clojure-lsp/clojure-lsp/issues/266).

I also noticed that calling `reformat-string` on the same range of `reformat-form` is faster.

Example:
calling `reformat-string` on [this](https://github.com/clojure-lsp/clojure-lsp/blob/master/src/clojure_lsp/handlers.clj) whole buffer, takes __~450ms__ while calling `reformat-form` on the same range (the whole buffer) will take __~2000ms__.

lread:
Hi @ericdallo, I do realize you are compiling clojure-lsp to a binary executable with GraalVM's `native-image`, but I'll start exploring with regular JVM, namely AdoptOpenJDK v1.10.11 HotSpot.

I'm not sure the best way to approach this, but it seems sensible to first try to reproduce the performance issue you are encountering.  

My dev box is an aging but capable late 2013 iMac, with 32gb of RAM and a 3.5 GHz Quad-Core i7 and SSD storage.

I am exploring under the cljfmt dir on cljfmt master.

I've added a `deps.edn`:
```Clojure
{:paths ["src" "resources"]
 :deps {org.clojure/clojure {:mvn/version "1.10.3"}
        rewrite-clj/rewrite-clj {:mvn/version "1.0.605-alpha" }
        criterium/criterium {:mvn/version "0.4.6"}}}
```

And a `perf.clj` to hopefully mimic your 2 scenarios:
```Clojure
(ns perf
  (:require [cljfmt.core :as cljfmt]
            [criterium.core :as criterium]
            [rewrite-clj.parser :as p]))

(def sample1 "https://raw.githubusercontent.com/clojure-lsp/clojure-lsp/cd64d430d83cb068d4a0a011beeded3beb791201/src/clojure_lsp/handlers.clj")

(defn -main [[test-id]]
  (case test-id
    "reformat-string"
[...]

PR #234, opened 2021-05-24
Fix slow performance due to excessive calls to `find-namespace`

LuisThiamNye:
Ensure the `find-namespace` function is only called once per reformat request. Otherwise, it is called many times per line, significantly slowing down the formatter.

Addresses #226

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -220,9 +220,6 @@
 (defn- ns-form? [zloc]
   (some-> zloc z/down ns-token?))
 
-(defn- find-namespace [zloc]
-  (some-> zloc top (z/find z/next ns-form?) z/down z/next z/sexpr))
-
 (defn- indent-matches? [key sym]
   (if (symbol? sym)
     (cond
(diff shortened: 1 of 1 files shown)


cljfmt/src/cljfmt/core.cljc
```diff
@@ -89,7 +89,7 @@
   (z/skip z/next* space? zloc))
 
 (defn- skip-whitespace-and-commas [zloc]
-  (z/skip z/next* #(or (space? %) (comma? %)) zloc)) 
+  (z/skip z/next* #(or (space? %) (comma? %)) zloc))
```
weavejester, line 92: Could you snip out the whitespace fixes from the commit?

cljfmt/src/cljfmt/core.cljc
```diff
@@ -293,54 +290,58 @@
...
   ([form]
-   (indent form default-indents))
+   (indent form default-indents {}))
   ([form indents]
-   (transform form edit-all should-indent? #(indent-line % indents {})))
+   (indent form indents {}))
   ([form indents alias-map]
-   (transform form edit-all should-indent? #(indent-line % indents alias-map))))
+   (let [ns-name (find-namespace (z/edn form))]
+     (transform form edit-all should-indent? #(indent-line % indents {:alias-map alias-map :ns-name ns-name})))))
```
weavejester, line 344: Can you limit the length of this line to 80 characters?

weavejester (changes requested):
Thanks for the patch, and apologies about the delay getting to it. Just have a couple of minor formatting issues and then I can merge it in.

PR #250, opened 2022-04-27
Improve heuristics for finding ns form

### LINKED ISSUE #247: reformat-form throws when a local variable is named "ns"
Trying to call `reformat-form` when a local variable is named `ns` throws an Exception

```clojure
(let [ns (range 10)]
  (reduce + ns
          ))
```

<details>
  <summary>
    Exception
  </summary>
<pre><code>
                                         cljfmt.core/reformat-form                          core.cljc:  416
                                              cljfmt.core/reindent                          core.cljc:  379
                                                cljfmt.core/indent                          core.cljc:  346
                                                               ...
                                             cljfmt.core/transform                          core.cljc:   33
                                             cljfmt.core/transform                          core.cljc:   34
                                                clojure.core/apply                           core.clj:  669
                                                               ...
                                              cljfmt.core/edit-all                          core.cljc:   30
                                             cljfmt.core/indent/fn                          core.cljc:  347
                                           cljfmt.core/indent-line                          core.cljc:  331
                                         cljfmt.core/indent-amount                          core.cljc:  326
                                         cljfmt.core/custom-indent                          core.cljc:  318
                                          clojure.core/some-fn/spn                           core.clj: 7536
                                                 clojure.core/some                           core.clj: 2718
                                       clojure.core/some-fn/spn/fn                           core.clj: 7538
                                          clojure.core/some-fn/sp1                           core.clj: 7513
                                                 cljfmt.core/fn/fn                          core.cljc:  298
                                          cljfmt.core/inner-indent                          core.cljc:  261
                                cljfmt.core/fully-qualified-symbol                          core.cljc:  256
                             cljfmt.core/qualify-symbol-by-ns-name                          core.cljc:  250
                                                 clojure.core/name                           core.clj: 1610
java.lang.ClassCastException: class clojure.lang.PersistentList cannot be cast to class clojure.lang.Named (clojure.lang.PersistentList and clojure.lang.Named are in unnamed module of loader 'app')
</code></pre>
</details>

I would expect it to reformat the same as it does when the local variable is named something else:

```clojure
(let [ms (range 10)]
  (reduce + ms
          ))

;; becomes

(let [ms (range 10)]
  (reduce + ms))
```

mainej:
The helper that finds the namespace form was too broad in what it would accept.

For example, if the first `ns` token it found looked like the following, it would consider it an ns form:

```clojure
(let [ns (range 100)] ,,,)
```

In this case an exception was thrown because cljfmt tries to treat `(range 100)` as the ns name, and calls `clojure.core/name` on it.

Even if the next expression responded to `clojure.core/name`, weird things could happen (see tests) because cljfmt would use the wrong ns name:

```clojure
(let [ns another-local-var] ,,,)
```

This commit makes the heuristics for finding a ns form more restrictive. Now, an ns form must be ~a list that starts with at least two tokens, the token ns and another token following it~ a top-level list that starts with the `ns` token.

```clojure
(ns my-ns ,,,)
```

I may run into this more than others because I work on clojure-lsp and a little on clj-kondo, and so it's common to have local vars named ns. Also, clojure-lsp supports formatting of individual defns, separate from the rest of their namespace, which means the true ns form is missing.

Fixes #247.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -41,16 +41,16 @@
   (nil? (z/up* zloc)))
 
 (defn- top? [zloc]
-  (and zloc (not= (z/node zloc) (z/root zloc))))
+  (some-> zloc z/up root?))
 
-(defn- top [zloc]
+(defn- root [zloc]
   (if (root? zloc) zloc (recur (z/up zloc))))
 
 (defn- clojure-whitespace? [zloc]
   (z/whitespace? zloc))
 
 (defn- surrounding-whitespace? [zloc]
-  (and (top? (z/up zloc))
+  (and (not (top? zloc))
        (surrounding? zloc clojure-whitespace?)))
 
 (defn remove-surrounding-whitespace [form]
@@ -220,7 +220,9 @@
        (= 'ns (z/sexpr zloc))))
 
 (defn- ns-form? [zloc]
-  (some-> zloc z/down ns-token?))
+  (and (top? zloc)
+       (= (z/tag zloc) :list)
+       (some-> zloc z/down ns-token?)))
 
 (defn- indent-matches? [key sym]
   (if (symbol? sym)
@@ -334,7 +336,7 @@
       zloc)))
 
 (defn- find-namespace [zloc]
-  (some-> zloc top (z/find z/next ns-form?) z/down z/next z/sexpr))
+  (some-> zloc root (z/find z/next ns-form?) z/down z/next z/sexpr))
 
 (defn indent
   ([form]
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -336,7 +336,7 @@
       zloc)))
 
 (defn- find-namespace [zloc]
-  (some-> zloc root (z/find z/next ns-form?) z/down z/next z/sexpr))
+  (some-> zloc root z/down (z/find z/right ns-form?) z/down z/next z/sexpr))
 
 (defn indent
   ([form]
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -281,7 +281,33 @@
           "                  (prn x)))"]
          {:indents {'thing.core/defrecord [[:inner 0]]}
           #?@(:cljs [:alias-map {"t" "thing.core"}])})
-        "applies custom indentation to namespaced defrecord"))
+        "applies custom indentation to namespaced defrecord")
+    (is (reformats-to?
+         ["(let [ns subs]"
+          "  (ns \"string\"))"
+          ""
+          "(ns thing.core)"
+          ""
+          "(defthing foo [x]"
+          "(+ x 1))"]
+         ["(let [ns subs]"
+          "  (ns \"string\"))"
+          ""
+          "(ns thing.core)"
+          ""
+          "(defthing foo [x]"
+          "  (+ x 1))"]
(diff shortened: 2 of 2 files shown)


mainej:
> Is there a reason you didn't choose to check if the ns form is top-level

No, you're right, that seems better. In practice, I've never seen an ns form that isn't top-level.

mainej:
But @weavejester, that brings up another question... `top?` seems buggy:

```clojure
(-> (z/of-string "(let [ns a] (ns a))")
    (z/down)
    (z/right)
    (z/right)
    (z/down) ;; second `ns` token
    top?)
;; => true
```

Is that accidental? Would you like this PR to include a fix for `top?`?

mainej:
I have a fix for `top?` which returns true only for top-level zlocs, instead of all non-root zlocs, which is what it did before. That fix required an adjustment to `surrounding-whitespace?` which was depending on the buggy behavior of `top?`. I'm going to (force) push that fix, but let me know if you'd like it split into a separate PR.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -41,16 +41,16 @@
   (nil? (z/up* zloc)))
 
 (defn- top? [zloc]
-  (and zloc (not= (z/node zloc) (z/root zloc))))
+  (some-> zloc z/up root?))
 
-(defn- top [zloc]
+(defn- root [zloc]
```
mainej, line 46: This rename reflects that this helper navigates to the root zloc, which is one level above the top-level zlocs.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -41,16 +41,16 @@
...
   (if (root? zloc) zloc (recur (z/up zloc))))
 
 (defn- clojure-whitespace? [zloc]
   (z/whitespace? zloc))
 
 (defn- surrounding-whitespace? [zloc]
-  (and (top? (z/up zloc))
+  (and (not (top? zloc))
```
mainej, line 53: This was depending on the buggy definition of `top?`, which if it had been named according to it's implementation would have been called `non-root?`. The refactoring that arrived at this form went like this:
```clojure
(top? (z/up zloc))       ;; original, with buggy definition of top?
(non-root? (z/up zloc))  ;; rename top? to reflect actual behavior
(not (root? (z/up zloc)) ;; replace with negation of converse
(not (top? zloc))        ;; final, with new correct definition of top?
```

cljfmt/src/cljfmt/core.cljc
```diff
@@ -334,7 +336,7 @@
       zloc)))
 
 (defn- find-namespace [zloc]
-  (some-> zloc top (z/find z/next ns-form?) z/down z/next z/sexpr))
+  (some-> zloc root z/down (z/find z/right ns-form?) z/down z/next z/sexpr))
```
mainej, line 339: Since `ns-form?` is only true for top-level forms, it's more efficient to search `z/right` across the top level forms, rather than `z/next`, down through all the forms.

mainej, line 339: Since `ns-form?` is only true for top-level forms, it's more efficient to search `z/right` across the top level forms, rather than `z/next`, down through all the forms.

weavejester:
Thanks, this looks good. Can you change the first commit message to:

```
Improve heuristics for finding ns form

The ns-form? function is too broad in what it accepts. For example, it
considers this to be an valid ns form:

  (let [ns (range 100)] ...)

This commit makes the function more restrictive: an ns form must now be
a top-level list that starts with the symbol ns.

Fixes #247.
```

It's a little more concise without losing information.

mainej:
> a little more concise

👍 Much more concise! Done.

PR #252, opened 2022-05-01
Address clj-kondo warnings

or:
I hope it's ok if these are together in one PR. The commits are relatively small:

- minor refactorings to adhere to https://github.com/bbatsov/clojure-style-guide and address clj-kondo warnings
- ~`cond->` merging~
- ~refactoring to keep cljfmt options and their defaults in only one place~

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -225,13 +225,13 @@
        (some-> zloc z/down ns-token?)))
 
 (defn- indent-matches? [key sym]
-  (if (symbol? sym)
+  (when (symbol? sym)
     (cond
       (symbol? key)  (= key sym)
       (pattern? key) (re-find key (str sym)))))
 
 (defn- token-value [zloc]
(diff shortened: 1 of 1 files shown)


weavejester:
I think these would be better separated out further, as these its easier to keep track of what to accept and what to reject when they're individual PRs.

or:
Alright! If you don't mind many small PRs, then I'm all for it. I'll do 3, based on the bullet points.

or:
The third one relies on the second, so I can't create it yet without duplicating those commits. I'll wait for feedback.

weavejester:
Can you remove the "Prefix unused bindings" commit? It's not necessary for readability.

or:
Done, but it is something clj-kondo warns about.

weavejester:
> Done, but it is something clj-kondo warns about.

Sure, but we shouldn't blindly follow what a style linter says if it achieves the opposite of what we want.

Can you also change the commit to:

```
Use 'when' instead of 'if' for single branches

Recommended by the Clojure Style Guide.
```

or:
Done.

PR #253, opened 2022-05-01
Merge multiple 'cond->' expressions into one

or:
This simplifies the expression quite a bit and is equivalent because
'cond->' doesn't short circuit after true tests.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -404,22 +404,27 @@
   ([form]
    (reformat-form form {}))
   ([form opts]
-   (-> form
-       (cond-> (:split-keypairs-over-multiple-lines? opts false)
-         (split-keypairs-over-multiple-lines))
-       (cond-> (:remove-consecutive-blank-lines? opts true)
-         remove-consecutive-blank-lines)
-       (cond-> (:remove-surrounding-whitespace? opts true)
-         remove-surrounding-whitespace)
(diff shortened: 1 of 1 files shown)


weavejester:
Is there a performance benefit to this?

or:
I haven't investigate it, but I doubt it.
I don't think using `cond->` like that is idiomatic, it also adds complexity to the form, no?
Is there an advantage in keeping multiple `cond->`s that I'm not aware of?

weavejester:
The indentation makes it easier to determine condition and result without resorting to vertical whitespace.

or:
Hmm, I guess I can see that. I feel like the vertical whitespace is optional, but also less intrusive than the extra form.

I personally try to avoid putting a form on the same line as nested threading macros, because at first glance it looks like the first argument is the one being threaded, e.g.:
```clj
(-> foobar
     (cond-> A
        B))
```
It's not immediately clear that `A` isn't the first argument there. At least not to me, but that's mostly personal preference.
(An exception is `as->`, but that also works differently, so it's an immediate visual distinction.)

Feel free to close the PR if you're not convinced, it's not a big deal. :)

weavejester:
I think I'll keep it how it is, but you do make a good argument. I think in this case using `cond->` with `->` is fine, as that was the intended use case when it was developed. Avoiding overuse of vertical whitespace is also important, as it uses valuable screen real estate.

Issue #254, opened 2022-05-02
Improve performance


or:
As most users, I use linters and formatters mainly in my editor, but also to check or fix a larger code base at the command line or in workflows.
For both of those usages performance is a factor, so I've been trying to reduce `cljfmt` start-up and run time.

- I've experimented with using `pmap` to process multiple files for `fix`, and that sped up fixing a largeish directory by a factor of 2-3, so I think a `--parallel` option would be useful, would you consider a PR for that feature?
- the plugin to create a native image currently has a bug that prevents the `:jvm-opts` specified for it to be used (see https://github.com/taylorwood/lein-native-image/pull/8), if those flags are used, then the executable is about 20% faster on my machine (with GraalVM CE or EE); I only mention this here as a heads up, as I hope the plugin will be fixed soon, but if that takes longer than expected it might be worth to use the workaround
- a small performance improvement for native-image is not using `read-string` here: https://github.com/or/cljfmt/blob/da37b7aef08be99dbb8e2e8772d95308150d7adb/cljfmt/src/cljfmt/main.clj#L146, which is done to allow `.clj` files for indents and aliases; a little surprisingly this doubles the size of the native-image executable and removing it improves performance by about 4-5% - I realize that's not huge, but would you consider a build flag that turns off .clj support?

The largest improvement is using GraalVM EE, which has nothing to do with `cljfmt`, but it almost doubles the speed out of the box, and even there the `jvm-opts` and clj-support removal have a noticeable impact. For the largest CLJS file in my project the current master took a median of 556ms to fix an already formatted file (with the various options I use), with the changes it's down to 443ms. With GraalVM EE this crumbles to 258ms.
These are all acceptable numbers, but I think it's worth optimizing this as much as possible, so it becomes unnoticeable when used in an on-save handler in the editor.

For reference:
<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">

<colgroup>
<col  class="org-left" />

<col  class="org-left" />

<col  class="org-left" />

<col  class="org-right" />

<col  class="org-right" />

<col  class="org-right" />

<col  class="org-right" />

<col  class="org-left" />
</colgroup>
<thead>
<tr>
<th scope="col" class="org-left">version</th>
<th scope="col" class="org-left">GraalVM</th>
<th scope="col" class="org-left">size in MB</th>
<th scope="col" class="org-right">&#x2013;help</th>
<th scope="col" class="org-right">check in ms</th>
[...]

PR #258, opened 2022-05-07
Refactor indenting

or:
This doubles the speed for multi file processing in a native image. By separating the identification of parent form-symbols from the rule matching and using map look-ups.

Related to #254.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -260,7 +260,7 @@
 
 (defn- inner-indent [zloc key depth idx context]
   (let [top (nth (iterate z/up zloc) depth)]
-    (when (and (or (indent-matches? key (fully-qualified-symbol zloc context))
+    (when (and (or (indent-matches? key (fully-qualified-symbol top context))
                    (indent-matches? key (remove-namespace (form-symbol top))))
                (or (nil? idx) (index-matches-top-argument? zloc depth idx)))
       (let [zup (z/up zloc)]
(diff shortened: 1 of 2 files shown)


weavejester:
Can you provide benchmarks? Does it improve the speed in Java as well?

In its current state, I'm afraid that the code isn't readable enough to be merged. Indexing the indentation rules in some way is a good idea, but we need to take care not to increase complexity or negatively impact maintainability. Can you give an overview of your design?

or:
The commit message includes concrete performance improvements, including the Java version. In short: it's not as big an impact, I'm thinking due to JIT, but still decent. It's hard to benchmark this with Java in particular, due to JIT and concurrency, but also because it's not easy to separate start-up time from processing time. For instance, the directory improvement given in the commit message probably underestimates the real impact substantially considering the measured times for a single file, which implies a ~2s start-up time.

I did a bunch of tests in Java and GraalVM with https://github.com/Tyruiop/needle and https://github.com/clojure-goes-fast/clj-async-profiler, focusing on number of calls and unnecessary work duplication.

The commit message also outlines the design a bit. I'm assuming you looked mostly at the diff yesternight, so let me know if the commit message already helps, otherwise I can write up some more details.

I *think* it already simplifies some aspects of the design, like `indenter-fn` and the main flow starting at `custom-indent`. But `build-indent-rule-map` and `applicable-indent-rules` definitely are a bit unwieldy. That's where most of the complexity lives, but the good news is that they also encapsulate the look-up logic.

PR #259, opened 2022-05-08
Refactor cljfmt options

or:
The goal is to simplify option handling and reduce duplication. It also revealed a minor bug in the test suite.

Side note: I noticed that the merge of default indents and specified indents is only done in main, so using `cljfmt.core/reformat-form` programmatically will overwrite `:indents` entirely. Is that by design? I guess it makes sense to be able to start with a clean slate, but it could be surprising.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -400,26 +400,37 @@
 (defn remove-multiple-non-indenting-spaces [form]
   (transform form edit-all non-indenting-whitespace? replace-with-one-space))
 
+(def default-options
+  {:indentation?                          true
+   :insert-missing-whitespace?            true
+   :remove-consecutive-blank-lines?       true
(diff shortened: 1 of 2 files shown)


weavejester:
Thanks for the PR. Can you move the unit test fix out, as it's not part of the cljfmt options as far as I can see. Can you also remove the commit that adds the options programmatically, as I think it reduces readability.

or:
Done. Tests will pass once https://github.com/weavejester/cljfmt/pull/261 is merged.

weavejester:
> Side note: I noticed that the merge of default indents and specified indents is only done in main, so using `cljfmt.core/reformat-form` programmatically will overwrite `:indents` entirely. Is that by design?

Yes. At the command line the defaults are merged for convenience, but the API needs to allow the developer to override the indents entirely. The CLI is "easier" while the API is "simpler".

PR #264, opened 2022-05-10
Fix native image build

or:
Fix the ignored `:jvm-opts` for native-image, as mentioned in #254.

Also clean up the native-image build, some options have no effect or didn't work as intended on current GraalVM versions (21.2.0+, latest is 22.1).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ pom.xml.asc
 *.class
 .lein-*
 .nrepl-port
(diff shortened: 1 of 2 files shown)


weavejester:
Thanks - can you squash the "remove option" commits together? So 64236e0, 3ef275c and 5167941.

or:
Yup, done.

Issue #256, opened 2022-05-02
Feature request: a configuration file


or:
Having the aliases, indents, and formatting options passed as CLI arguments is not ideal for editor integration and using `cljfmt` in multiple projects.

It'd be awesome if `cljfmt` could read all options from a single file, which can be specified. By default it could look for `.cljfmt.edn` (or similar) in the current directory and each directory up to HOME.
It should use the first file it finds, without any attempt to merge with files higher up the directory tree, so there aren't any surprises due to a user's personal preferences.

This would allow per-project configuration for projects that don't use Leiningen and simplify editor/workflow integration.

or:
Thanks, that's good to know. Unfortunately that won't help me, because I use the cljfmt binary for formatting, which also can check/fix whole directories.

I think it would be better if cljfmt read its own configuration file, then LSP probably wouldn't have to implement it.

weavejester:
I'd be open to a PR that would look for a configuration file in the current directory. I'm less sure about looking in HOME or checking parent directories.

or:
Using the command line tool for linting in Emacs will run it in the file's directory, not the project root, so it needs to find the configuration.

And having a global default in one place (HOME) is incredibly useful as well. It would be painful to have to copy (and maintain) files in various project directories.

sirmspencer:
Would this extend to the lein-cljfmt plugin, or would that need to be a seperate change on that side?

weavejester:
> Would this extend to the lein-cljfmt plugin, or would that need to be a seperate change on that side?

No, because Leiningen already has per-project and per-user configurations.

PR #274, opened 2022-10-16
Add configuration file to CLI version

or:
Similar to the Leiningen configuration it would be useful to put rules in a configuration file, which can then be version controlled and honoured by tools that only run cljfmt in the directory, without taking command line arguments.

The file .cljfmt.edn is looked for in the current directory and all its parent directories, the first file found is used.

Command line options override the options in the configuration file, command line indents are added to the default indents or those found inside the configuration file.

The configuration file can also be specified at the command line directly to override looking for it.

This addresses #256.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,7 @@ clojure -Sdeps '{:deps {cljfmt {:mvn/version "0.9.0"}}}' \
 
 ## Configuration
 
+### Leiningen
 You can configure lein-cljfmt by adding a `:cljfmt` map to your
 project:
 
@@ -108,6 +109,13 @@ project:
 :cljfmt {}
 ```
 
+### Command line
+You can configure `cljfmt` by placing a `.cljfmt.edn` or `.cljfmt.clj`
+file in the directory where you run the command, or any parent
+directory. The first file in the hierarchy will be used, files in
+higher up parent directories are ignored.
+
+### Rules
 cljfmt has several different formatting rules, and these can be
 selectively enabled or disabled:
 
--- a/cljfmt/src/cljfmt/main.clj
+++ b/cljfmt/src/cljfmt/main.clj
@@ -1,12 +1,12 @@
 (ns cljfmt.main
   "Functionality to apply formatting to a given project."
   (:require [cljfmt.core :as cljfmt]
-            [clojure.string :as str]
+            [cljfmt.diff :as diff]
             [clojure.edn :as edn]
             [clojure.java.io :as io]
             [clojure.stacktrace :as st]
-            [clojure.tools.cli :as cli]
-            [cljfmt.diff :as diff])
+            [clojure.string :as str]
+            [clojure.tools.cli :as cli])
   (:gen-class))
 
 (defn- abort [& msg]
@@ -154,68 +154,89 @@
       (read-string contents)
       (edn/read-string contents))))
 
-(def default-options
+(defn- parent-dirs [^String root]
+  (->> (.getAbsoluteFile (io/file root))
+       (iterate #(.getParentFile ^java.io.File %))
+       (take-while some?)))
+
+(defn- find-file-in-dir ^java.io.File [^java.io.File dir ^String name]
+  (let [f (io/file dir name)]
+    (when (.exists f) f)))
+
+(defn- find-config-file-in-dir ^java.io.File [^java.io.File dir]
+  (or (find-file-in-dir dir ".cljfmt.edn")
+      (find-file-in-dir dir ".cljfmt.clj")))
+
+(defn- load-configuration []
+  (some->> (parent-dirs "")
+           (some find-config-file-in-dir)
+           (#(.getPath ^java.io.File %))
+           cli-file-reader))
+
+(defn- load-default-options []
   (merge cljfmt/default-options
          {:project-root "."
           :file-pattern #"\.clj[csx]?$"
           :ansi?        true
-          :parallel?    false}))
+          :parallel?    false}
+         (load-configuration)))
 
-(defn merge-default-options [options]
-  (-> (merge default-options options)
-      (assoc :indents (merge (:indents default-options)
+(defn merge-default-options [defaults options]
+  (-> (merge defaults options)
+      (assoc :indents (merge (:indents defaults)
                              (:indents options {})))))
 
 (def default-paths ["src" "test" "project.clj"])
 
-(def ^:private cli-options
+(defn- cli-options [defaults]
   [[nil "--help"]
    [nil "--parallel"
     :id :parallel?]
    [nil "--project-root PROJECT_ROOT"
-    :default (:project-root default-options)]
+    :default (:project-root defaults)]
    [nil "--file-pattern FILE_PATTERN"
-    :default (:file-pattern default-options)
+    :default (:file-pattern defaults)
     :parse-fn re-pattern]
    [nil "--indents INDENTS_PATH"
     :parse-fn cli-file-reader]
    [nil "--alias-map ALIAS_MAP_PATH"
     :parse-fn cli-file-reader]
    [nil "--[no-]ansi"
-    :default (:ansi? default-options)
+    :default (:ansi? defaults)
     :id :ansi?]
    [nil "--[no-]indentation"
-    :default (:indentation? cljfmt/default-options)
+    :default (:indentation? defaults)
     :id :indentation?]
    [nil "--[no-]remove-multiple-non-indenting-spaces"
-    :default (:remove-multiple-non-indenting-spaces? cljfmt/default-options)
+    :default (:remove-multiple-non-indenting-spaces? defaults)
     :id :remove-multiple-non-indenting-spaces?]
    [nil "--[no-]remove-surrounding-whitespace"
-    :default (:remove-surrounding-whitespace? cljfmt/default-options)
+    :default (:remove-surrounding-whitespace? defaults)
     :id :remove-surrounding-whitespace?]
    [nil "--[no-]remove-trailing-whitespace"
-    :default (:remove-trailing-whitespace? cljfmt/default-options)
+    :default (:remove-trailing-whitespace? defaults)
     :id :remove-trailing-whitespace?]
    [nil "--[no-]insert-missing-whitespace"
-    :default (:insert-missing-whitespace? cljfmt/default-options)
+    :default (:insert-missing-whitespace? defaults)
     :id :insert-missing-whitespace?]
    [nil "--[no-]remove-consecutive-blank-lines"
-    :default (:remove-consecutive-blank-lines? cljfmt/default-options)
+    :default (:remove-consecutive-blank-lines? defaults)
     :id :remove-consecutive-blank-lines?]
    [nil "--[no-]split-keypairs-over-multiple-lines"
-    :default (:split-keypairs-over-multiple-lines? cljfmt/default-options)
+    :default (:split-keypairs-over-multiple-lines? defaults)
     :id :split-keypairs-over-multiple-lines?]
    [nil "--[no-]sort-ns-references"
-    :default (:sort-ns-references? cljfmt/default-options)
+    :default (:sort-ns-references? defaults)
     :id :sort-ns-references?]])
 
 (defn- file-exists? [path]
   (.exists (io/as-file path)))
 
 (defn -main [& args]
-  (let [parsed-opts   (cli/parse-opts args cli-options)
+  (let [defaults      (load-default-options)
+        parsed-opts   (cli/parse-opts args (cli-options defaults))
         [cmd & paths] (:arguments parsed-opts)
-        options       (merge-default-options (:options parsed-opts))
+        options       (merge-default-options defaults (:options parsed-opts))
         paths         (or (seq paths) (filter file-exists? default-paths))]
     (if (:errors parsed-opts)
       (abort (:errors parsed-opts))


weavejester:
Thanks for the PR. As I mentioned in the original issue, don't think it should look for `.cljfmt.edn` files in parent directories, just in the current directory. Traversing up the tree has potential for producing unexpected behaviour.

I'm not sure what the `DefaultValue` record is for, or why the additional complexity is needed. Can you explain why you came up with the design you did, and why you didn't just check for a file and merge it in if it exists?

or:
> Thanks for the PR. As I mentioned in the original issue, don't think it should look for `.cljfmt.edn` files in parent directories, just in the current directory. Traversing up the tree has potential for producing unexpected behaviour.

I think it has to do this to be useful for editors, like emacs, which run the formatter/linter in the directory where the file is by default. Maybe this can be tweaked somehow, but it would complicate matters a lot. All other linters I can currently think of behave like that.
I guess it should gracefully ignore access errors when trying to test the existence of files or opening them, but are there other problems you are thinking of?

> I'm not sure what the `DefaultValue` record is for, or why the additional complexity is needed. Can you explain why you came up with the design you did, and why you didn't just check for a file and merge it in if it exists?

Sure, I meant to elaborate on that, but had to step away earlier. I believe generally the arguments passed to the command take highest precedence, so they should override anything in the file. So merging "defaults", then "file", then "args" makes sense. However, the options parsed via tools.cli are just the effective values, which are the defaults if the user didn't specify anything, so they'd just overwrite everything again.

The idea was to keep track of what is a default value and what was specified by the user, and only the latter args are merged in.

weavejester:
Editors should know the project directory, but I'd be willing to reconsider if I have some use-case examples that demonstrate that looking for a configuration directory up the tree is going to make things easier.

Regarding the default values, I think it makes more sense to use the defaults from the configuration. That way we don't need to use a record, and the `--help` message will show the correct defaults derived from the configuration (if one exists).

Finally, I think we can make the configuration finding function a little more concise:

```clojure
(defn find-config-file [filename]
  (loop [dir (.getAbsoluteFile. (File. "."))]
    (when dir
      (let [file (File. dir filename)]
        (if (.exists file)
          file
          (recur (.getParentFile dir)))))))

**Comment by or** [comment]
Changed to the loop.

As for the editors: not every editor has a concept of "project", some open files just with a path. And even with something like `projectile` Emacs uses the file's directory. I haven't tested it with vim or others, but I believe this is very common.

**Comment by or** [comment]
> Regarding the default values, I think it makes more sense to use the defaults from the configuration. That way we don't need to use a record, and the `--help` message will show the correct defaults derived from the configuration (if one exists).

Yes, this is a good idea, and I considered it as well. The default values are misleading once they are printed out (e.g. after https://github.com/weavejester/cljfmt/pull/263). It should say something like "loaded from <path-to-file>" and then show those values.

Alas, since `--config-file` can still be passed along, it's not that easy. I think there'd have to be a tools.cli setup first to parse `--config-file`, then use that information to read the right configuration file, then use tools.cli again. Tricky.
I can draft something like that if you think that's ok, maybe it isn't as messy as I expect it to be.

I think the user experience would indeed be much better, and `DefaultValue` could be thrown out.

**Comment by weavejester** [comment]
Why have `--config-file` at all? Is there a use-case for it?

**Comment by or** [comment]
Good question, that's probably an optional part of the feature.

I think most other formatters/linters have it. Perhaps to address cases where the search for a config file doesn't work. E.g. the code is somewhere in `/projects/foobar`, but a shared configuration is kept in `/etc/cljfmt.edn`, or it is known that the process will not have access to some of the directories the traversal would check, so specifying it explicitly is necessary.

I can drop it to keep things simple for now, and the feature can be requested when it is required.

**Comment by weavejester** [comment]
Let's drop it for now, and if it's an issue it can be part of a PR in the future.

**Comment by mainej** [comment]
I have two requests related to this PR. It'd be great if they could be included in it, but if not, perhaps they can be added soon?
1. If possible, I'd love to have these new tools be public, the tools that read the `.cljfmt.(edn|clj)` file and merge in default opts. Other libraries like [Calva](https://github.com/clojure-lsp/clojure-lsp/blob/d774e00c3d777bb5ededcc779310758c18262c3d/lib/src/clojure_lsp/feature/format.clj#L20-L32) and [clojure-lsp](https://github.com/clojure-lsp/clojure-lsp/blob/master/lib/src/clojure_lsp/feature/format.clj#L20-L32) are currently reading and parsing these files themselves, but it would be nice to refactor them to call helpers in `cljfmt.main`.
2. Out of the box EDN doesn't support regular expressions. So there's no way to express the .clj syntax `:indents {#"^for-" [[:inner 0]]}` in an EDN file. That is, at the moment it isn't possible to define indents that use regular expressions in EDN files. There is a workaround however, pioneered by clojure-lsp—you can [define](https://github.com/clojure-lsp/clojure-lsp/blob/d774e00c3d777bb5ededcc779310758c18262c3d/lib/src/clojure_lsp/feature/format.clj#L32) a reader function that interprets `#re "^for-"`, which is valid EDN, as a regular expression. So the above config can be written in an EDN file as `:indents {#re "^for-" [[:inner 0]]}`. It would be nice if cljfmt supported regular expressions in EDN files with this trick.

**Comment by mainej** [comment]
EDIT: I didn't read the new code quite carefully enough. At the moment it only reads `.cljfmt.edn`. It would be nice if it also tries `.cljfmt.clj`. It looks like `cli-file-reader` is already smart enough to handle either kind of file.

File: cljfmt/src/cljfmt/main.clj [hunk 1]
```diff
@@ -154,71 +154,91 @@
...
+    (when dir
+      (let [file (java.io.File. dir filename)]
+        (if (.exists file)
+          (.getAbsolutePath file)
+          (recur (.getParentFile dir)))))))
+
+(defn- load-configuration []
+  (when-let [config-file (find-config-file ".cljfmt.edn")]
+    {:config-file config-file
+     :config-data (cli-file-reader config-file)}))
```
weavejester, line 168: I don't think it's necessary to keep track of the file in this way. Let's keep it simple and just get the configuration.

cljfmt/src/cljfmt/main.clj
```diff
@@ -154,71 +154,91 @@
...
+
+(defn- load-defaults []
```
weavejester, line 170: Let's call this `load-default-options`.

README.md
```diff
@@ -101,13 +101,20 @@ clojure -Sdeps '{:deps {cljfmt {:mvn/version "0.9.0"}}}' \
...
+### Leiningen
 You can configure lein-cljfmt by adding a `:cljfmt` map to your
 project:
 
 ```clojure
 :cljfmt {}
 ```
 
+### Command line
+You can configure `cljfmt` by placing a `.cljfmt.edn` file in the directory
```
weavejester, line 113: Lines should be 72 characters or fewer the the README.

README.md
```diff
@@ -225,7 +232,7 @@ Indentation types are:
 * `:inner` -
   two character indentation applied to form arguments at a depth
   relative to a form symbol
-
```
weavejester, line 228: Remove any formatting corrections not relevant to the PR. It makes the diff much easier to read.

weavejester (changes requested):
Apologies for the delay in reviewing this. It looks good, and only needs minor changes (detailed in the comments).

README.md
or: done

README.md
or: done

cljfmt/src/cljfmt/main.clj
or: done

cljfmt/src/cljfmt/main.clj
or: done

or:
Also added `.cljfmt.clj` support.

cljfmt/src/cljfmt/main.clj
```diff
@@ -154,71 +154,88 @@
...
+      (let [file1 (java.io.File. dir ".cljfmt.edn")
+            file2 (java.io.File. dir ".cljfmt.clj")]
+        (cond
+          (.exists file1) (.getAbsolutePath file1)
+          (.exists file2) (.getAbsolutePath file2)
+          :else (recur (.getParentFile dir)))))))
+
+(defn- load-configuration []
+  (some-> (find-config-file)
+          cli-file-reader))
```
weavejester, line 169: This would be better written as:

```clojure
(defn find-config-file [filename]
  (loop [dir (.getAbsoluteFile. (io/file "."))]
    (when dir
      (let [file (io/file dir filename)]
        (if (.exists file)
          file
          (recur (.getParentFile dir)))))))

(defn- load-configuration []
  (some-> (or (find-config-file ".cljfmt.edn")
              (find-config-file ".cljfmt.clj"))
          cli-file-reader))
```

cljfmt/src/cljfmt/main.clj
```diff
@@ -154,71 +154,88 @@
...
+        options       (merge-default-options defaults (:options parsed-opts))
         paths         (or (seq paths) (filter file-exists? default-paths))]
     (if (:errors parsed-opts)
       (abort (:errors parsed-opts))
       (if (or (nil? cmd) (:help options))
-        (do (println "cljfmt [OPTIONS] COMMAND [PATHS ...]")
-            (println (:summary parsed-opts)))
+        (do
+          (println "cljfmt [OPTIONS] COMMAND [PATHS ...]")
+          (println (:summary parsed-opts)))
```
weavejester, line 238: Formatting change, should be reverted.

weavejester (changes requested):
Thanks for the update. There's a small formatting change to revert, and we can decomplect `find-config-file` by having it take an argument.

cljfmt/src/cljfmt/main.clj
or: done

cljfmt/src/cljfmt/main.clj
or: I considered that, but that'd find and use a `.cljfmt.edn` in `$HOME` (for instance) and ignore a closer `.cljfmt.clj` in the project directory.
I also considered passing both as arguments, but I think that makes `find-config-file` much more complicated and probably is not something that needs to be dynamic.

weavejester: Excellent point. In which case, perhaps:

```clojure
(defn- find-config-file [pred]
  (loop [dir (.getAbsoluteFile. (io/file "."))]
    (when dir
      (or (first (filter pred (.listFiles dir)))
          (recur (.getParentFile dir))))))

(def ^:private config-filenames #{".cljfmt.edn" ".cljfmt.clj"})

(defn- load-configuration []
  (some-> (find-config-file #(config-filenames (.getName %)))
          cli-file-reader))

Comment (or) [reply cljfmt/src/cljfmt/main.clj:169]: That's similar to my approach in passing the filenames as arguments, I also considered a loop over the filenames, but I still think both are more complicated without a real need.
It also seems inefficient, as it now needs to do a directory listing in every parent directory, which could be large or there might be permission issues. In particular when used as a formatter with an editor, this sort of thing can impact start-up time noticeably.

Since these filenames are pretty static, I think we should test them explicitly.

Comment (weavejester) [reply cljfmt/src/cljfmt/main.clj:169]: I generally don't like talking about efficiency without benchmarks, as it may be the case that even very large directories take little time to search through.

However, the permissions issues argument is a very good one. The process may have access to the file without having access to list the directory. So what about:

```clojure
(defn- parent-dirs [root]
  (->> (.getAbsoluteFile (io/file root))
       (iterate #(.getParentFile %))
       (take-while some?)))

(defn- find-file-in-dir [dir name]
  (let [f (io/file dir name)]
    (when (.exists f) f)))

(defn- find-config-file-in-dir [dir]
  (or (find-file-in-dir dir ".cljfmt.edn")
      (find-file-in-dir dir ".cljfmt.clj")

(defn- load-configuration []
  (some->> (parent-dirs ".")
           (some find-config-file-in-dir)
           cli-file-reader))
```

I believe your original design had a similar `parent-dirs` function?

or: Similar, yes. I used your version, added type hinting for GraalVM. The file also has to be converted to a string before `cli-file-reader` because there's an extension check there, which is a bit ugly, perhaps you have a better idea.
I think this likely was a problem in my previous iteration already. The branch I use personally doesn't have the `.clj` reading because it has a big performance impact.

cljfmt/src/cljfmt/main.clj
```diff
@@ -154,65 +154,86 @@
...
+    (when (.exists f) f)))
+
+(defn- find-config-file-in-dir ^java.io.File [^java.io.File dir]
+  (or (find-file-in-dir dir ".cljfmt.edn")
+      (find-file-in-dir dir ".cljfmt.clj")))
+
+(defn- load-configuration []
+  (some->> (parent-dirs "")
+           (some find-config-file-in-dir)
+           (#(.getPath ^java.io.File %))
```
weavejester, line 173: Couldn't this be `str`? Or we add a `str` to `filename-ext` in order to make that safe for files.

or: `str` works, but without the type hinting I get `java.io.File cannot be cast to java.lang.String` at runtime with GraalVM, so it'd look similar here. There could be an explicit type check in `filename-ext` as well, I guess.

weavejester:
I've merged some PRs that add a couple of conflicts to this one. Would you be able to resolve them? If not I should be able to find time this week or next to do so.

or:
Sure, done.

jakepearson:
I'm very excited about this. Need help getting in merged?

weavejester:
I've actually *just* started looking at the PRs that were opened at the end of last year, including this one. I've been on vacation over the holidays, and managed to catch COVID on the way back, so it's been a while since I've been able to review anything.

This all looks good. I've tested it and it seems fine. Let's just cut out the unnecessary parts of the commit message, and then it should be good to merge. Can you update the commit message to:

```
Add configuration file to CLI

The cljfmt.main function searches the current directory and all parent
directories for a configuration file named either .cljfmt.edn or
.cljfmt.clj. The first configuration file found is used.

Command line options override the options in the configuration file,
except for the indents, which are merged with the defaults.
```

or:
Done.

or:
And swift recovery!

weavejester:
Thanks!

sirmspencer:
Is this a different way to call from the command line vs the lein plug in?  If so, can you add an example to the readme?

weavejester:
> Is this a different way to call from the command line vs the lein plug in? If so, can you add an example to the readme?

https://github.com/weavejester/cljfmt#toolsdeps

or:
> Is this a different way to call from the command line vs the lein plug in? If so, can you add an example to the readme?

You can also compile it to a native binary via `lein native-image` if you get GraalVM: https://www.graalvm.org/java/quickstart/
Much better response times.

sirmspencer:
Thanks guys.

Issue #276, opened 2022-11-15
new option: auto expand aliased namespaced keywords 


mpenet:
Hi

I wanted to gauge if there's interest for a task that transforms all `::foo` type keywords into their fully expanded form `:current.namespace/foo`?

PR #277, opened 2022-11-15
[wip] initial implementation of expand nsed-keywords

mpenet:
As the title mentions, this auto-expand aliased keywords found to their full form, ::bar to :foo/bar, ::foo/bar to :something.foo/bar.

related to #276

I will clean it up & add tests if such an addition is considered valuable.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -350,6 +350,59 @@
                 #(indent-line % sorted-indents {:alias-map alias-map
                                                 :ns-name ns-name})))))
 
+(defn aliased-keyword? [zloc]
+  (let [node (z/node zloc)
+        ret (and (= :token (z/tag zloc))
+                 (n/keyword-node? node)
+                 (qualified-keyword? (n/sexpr node)))]
+    (and (= :token (z/tag zloc))
+         (n/keyword-node? node)
+         (qualified-keyword? (n/sexpr node)))))
+
+(defn- ns-require-form? [zloc]
+  (and (some-> zloc top-level-form ns-form?)
+       (some-> zloc z/child-sexprs first (= :require))))
+
+(defn- as-keyword? [zloc]
+  (and (= :token (z/tag zloc))
+       (= :as (z/sexpr zloc))))
+
+(defn- as-zloc->alias-mapping [as-zloc]
+  (let [alias (some-> as-zloc z/right z/sexpr)
+        current-namespace (some-> as-zloc z/leftmost z/sexpr)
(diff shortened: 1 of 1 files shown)


weavejester:
I'm not sure what this means. Can you give some examples?

weavejester:
Ah, I see the explanation is in the issue you opened, but I'm still not sure why this would be useful.

mpenet:
> Ah, I see the explanation is in the issue you opened, but I'm still not sure why this would be useful.

It's a way to undo keyword aliasing (`::...`}, it's can be considered (by some) to be bad practice and makes things more complicated to refactor among other things. I personally find it less readable when aliased.

weavejester:
My first thought on this is that making use of a feature in a language for its intended purpose isn't bad practice unless that feature has been deprecated. As cljfmt is intended to format Clojure code idiomatically, I don't believe this would under its scope.

mpenet:
Plenty of things are available and not encouraged to use though (refer :all comes to mind, or nested require path), some things are also personal preference (like alignment), so I don't think it's necessarily out of scope.

weavejester:
I think if we're going to say that namespace aliases for *functions* are idiomatic and good practice, then it's reasonable to say they're also idiomatic and good practice for keywords.

mpenet:
That's why I mentioned "by some". I don't claim it's the  only good/idiomatic choice, but it's one that's not uncommon in the wild and mentioned quite often. Then yes, it's a personal preference, I wouldn't advocate to make this the default, just like indentation/alignments, opinions vary. But I personally don't think it's unreasonable to consider this a valid option.

weavejester:
I think we have to draw the line somewhere. This is the first I've heard of this formatting choice, and you're the only person I know of who is advocating it. Any feature that's introduced to cljfmt needs to be reviewed and then maintained in perpetuity, and I don't think this feature quite makes the bar.

mpenet:
> I think we have to draw the line somewhere. This is the first I've heard of this formatting choice, and you're the only person I know of who is advocating it. Any feature that's introduced to cljfmt needs to be reviewed and then maintained in perpetuity, and I don't think this feature quite makes the bar.

I'd gladly maintain it :). If you look on clojurians there are more than a few people who mention this over time (hiredman, dpsutton, seancorfield & surely others, here's an example: https://app.slack.com/client/T03RZGPFR/C053AK3F9/thread/C03S1KBA2-1605313400.050700). But ultimately the decision is yours.

mpenet:
The code is fairly minimal, it mostly re-uses code that's already here. But I don't want to insist further, as you seem to have made your mind. 
Thanks

Issue #279, opened 2022-12-05
Do you want the `project.clj` file to be included by default for lein projects?


jacobemcken:
I am seeing this:
https://github.com/weavejester/cljfmt/blob/1003c677200e90bc19f7c8d83ea58bd17b900a5e/cljfmt/src/cljfmt/main.clj#L169

But the `project.clj` file is ignored when running `lein cljfmt check` (empty configuration), unless I specifically specify it (`lein cljfmt check project.clj`).

I've narrowed it down to `(seq paths)` never being empty when the paths come from the Leiningen plugin:
https://github.com/weavejester/cljfmt/blob/1003c677200e90bc19f7c8d83ea58bd17b900a5e/cljfmt/src/cljfmt/main.clj#L216

Do you want a PR changing `lein-cljfmt` to always include the `project.clj` file?

I was thinking that it could look something like (changing `lein-cljfmt/src/leiningen/cljfmt.clj`):
```patch
-           (filter #(and (.exists %) (.isDirectory %)))))))
+           (filter #(and (.exists %) (.isDirectory %)))
+           (concat '("project.clj"))))))
```

It seems to be a good fit since the project file is lein specific anyway.

weavejester:
This seems a reasonable change.

PR #280, opened 2022-12-10
Include project.clj by default for lein plugin

jacobemcken:
Since the file project.clj is specific to Leiningen projects it seem like a good default.

Resolves: #279


weavejester:
Thanks for the PR!

Why did you choose `concat` over `cons` here?

Also, can you change the commit message to:

```
Update Lein plugin to always include project.clj

Fixes #279.
```

I think the "by default" is misleading, because in this case it can't be changed. If this turns out to be an issue, we can add in an option to allow it to be omitted.

Have you tested to ensure that this plugin change works with both `check` and `fix`?

jacobemcken:
> Why did you choose `concat` over `cons` here?

My lack of knowledge.


> I think the "by default" is misleading, because in this case it can't be changed.

Either I don't understand meaning of your comment  or I disagree. As soon as you provide an alternative configuration in the `project.clj` file, like:
```clojure
...
    :cljfmt {:paths ["src"]}
...
```

The `project.clj` file is no longer include. What do you mean by "it can't be changed"?


I had only tested with `check` but now I have also successfully tested it with `fix` :+1: 
Thanks for the reminder.

weavejester:
> As soon as you provide an alternative configuration in the `project.clj` file, like:
> 
> ```clojure
> ...
>     :cljfmt {:paths ["src"]}
> ...
> ```
> 
> The `project.clj` file is no longer include

I don't think this is correct:

```clojure
leiningen.cljfmt=> (format-paths {:source-paths ["src"], :test-paths ["test"], :cljfmt {:paths ["src"]}})
("project.clj" #object[java.io.File 0x2ac11170 "src"])
```

You cons the string "project.clj" onto the list regardless of what is set in `:paths`. You've also put the string in the wrong place, as its not checked to see whether it exists (a small possibility, but a possibility), and it's not converted into a file instance.

jacobemcken:
> ... and it's not converted into a file instance.

The reason why it worked as a string regardless, is that both `check` and `fix` parses the list of "paths" through `find-files` which applies `io/file` again:
https://github.com/weavejester/cljfmt/blob/a4dd48ec94e9345b51eca13a086c91cef61a1fcd/cljfmt/src/cljfmt/main.clj#L33-L34

Thankfully, layering `io/file` on itself like `(io/file (io/file "some_file.txt"))` just returns a file object `#object[java.io.File 0x2ac11170 "some_file.txt"]`.

I've moved "extending `paths` with the project file" earlier, which also required changing how the filter works. An alternative would be to only check for `(.exists %)`. Is this what you had in mind?

PR #291, opened 2023-01-25
Combine `:alias-maps` setting with namespace aliases

bshepherdson:
Previously, if the `:alias-maps` setting was provided (even if it's an empty map) then only that map is used to look up indentation rules.

This means that `:indents` rules defined with a namespaced symbol only work if the var usage is fully qualified. In the common case of `(:require [foo.bar :as f])` cljfmt would not indent `f/something` according to the `:indents`.

With this change, the setting has the last word but the aliases in the file are not ignored.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -553,8 +553,8 @@
    (reformat-string form-string {}))
   ([form-string options]
    (let [parsed-form (p/parse-string-all form-string)
-         alias-map   #?(:clj (or (:alias-map options)
-                                 (alias-map-for-form parsed-form))
+         alias-map   #?(:clj (merge (alias-map-for-form parsed-form)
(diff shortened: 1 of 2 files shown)


weavejester:
Thanks for the fix! Can you add a test for it, so we can avoid regressions in future?

bshepherdson:
I've added a test. It's a little involved, but I think that covers the cases.

weavejester:
Thanks! Could you also update the commit message to:

```
Fix :alias-maps option to merge namespace aliases

Using the :alias-maps option should not replace aliases found by parsing
the namespace declaration.
```

(This project doesn't use markdown in its commit messages.)

bshepherdson:
Done.

bshepherdson:
@weavejester Any chance of landing this today? Anything else that needs doing?

weavejester:
Merged. Thanks for your work on this.

PR #295, opened 2023-02-02
Add babashka compatible badge to readme

lread:


weavejester:
While the intent is appreciated, I'd rather keep badges for CI statuses.

PR #297, opened 2023-02-02
Fix CLJS tests (fixes #296)

### LINKED ISSUE #296: CLJS tests are failing
Despite `lein test` returning a 0 exit code, it seems that the CLJS tests are failing.

Screenshot 2023-02-02 at 20 28 20

I'll follow up with a PR since I was the one who broke this, although I think the test runner should have exited with a non-zero exit code.

borkdude:
This fixes the tests like they were prior to #293. Fixes #296. 

This does _not_ address the problem that in case of a failing CLJS test, the exit code is non-zero since I want to keep this PR as minimal as possible.

### DIFF
--- a/cljfmt/bb.edn
+++ b/cljfmt/bb.edn
@@ -3,12 +3,9 @@
  {test {:doc "Run babashka tests"
         :extra-deps {eftest/eftest {:mvn/version "0.6.0"}}
         :extra-paths ["test"]
-        :requires ([eftest.runner :refer [find-tests run-tests]])
-        :task (let [{:keys [fail error]} (run-tests
-                                          (find-tests "test")
-                                          (when-not (System/console)
-                                            ;; better output in github actions
-                                            {:report clojure.test/report}))]
+        :requires ([clojure.test :as t]
+                   [cljfmt.core-test])
+        :task (let [{:keys [fail error]} (t/run-tests 'cljfmt.core-test)]
                 (when (or (pos? fail)
                           (pos? error))
                   (throw (ex-info "Tests failed" {:babashka/exit 1}))))}}}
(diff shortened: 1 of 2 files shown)


borkdude:
In case you want to migrate the CLJS tests to deps.edn, I have good experience with https://github.com/Olical/cljs-test-runner. That will also solve the exit code problem automatically. Just FYI.

cljfmt/bb.edn
```diff
@@ -3,12 +3,9 @@
...
         :extra-paths ["test"]
-        :requires ([eftest.runner :refer [find-tests run-tests]])
-        :task (let [{:keys [fail error]} (run-tests
-                                          (find-tests "test")
-                                          (when-not (System/console)
-                                            ;; better output in github actions
-                                            {:report clojure.test/report}))]
+        :requires ([clojure.test :as t]
+                   [cljfmt.core-test])
+        :task (let [{:keys [fail error]} (t/run-tests 'cljfmt.core-test)]
```
borkdude, line 8: I did not know how to avoid loading the `cljfmt.test-util.cljs` namespace with `eftest` so I switched to the vanilla clojure runner.

cljfmt/test/cljfmt/test_util/cljs.cljc
```diff
@@ -1,5 +1,5 @@
 (ns cljfmt.test-util.cljs
-  (:require [clojure.test :as test]
+  (:require [cljs.test :as test]
```
borkdude, line 2: It turned out that `cljs.test` is loaded here during macro-expansion, so it really needed to be `cljs.test` after all.

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -1,6 +1,5 @@
 (ns cljfmt.core-test
-  (:require [#?@(:clj (clojure.test :refer)
-                 :cljs (cljs.test :refer-macros)) [deftest testing is]]
+  (:require [clojure.test :refer [deftest testing is]]
```
borkdude, line 2: I can revert this change if you want, it slipped in, but it's no longer necessary to load `cljs.test` the way it was done prior to this PR.

PR #300, opened 2023-03-22
Add option to align maps and other forms

reutsharabani:
This commit adds the option to align maps and other forms. It does so by adding two separate configurable options:

 - align-maps?

 - align-forms?

Forms specifically are further configurable by a cljfmt option:

 - align-forms-spec

This is a map of the form `{form #{1 2}}` that maps the form's expected alignable positions. An example is `{let #{0}}` implying that a `let` symbol is immediately followed by an alignable form.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,17 @@ selectively enabled or disabled:
   other references in the `ns` forms at the top of your namespaces.
   Defaults to false.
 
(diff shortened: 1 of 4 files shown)


reutsharabani:
opened by mistake

Issue #301, opened 2023-04-02
`lein cljfmt check` ignores `.cljfmt.edn` configurations


raycheung:
Calling `lein cljfmt check` only reads configurations written in `project.clj`, not `.cljfmt.edn` / `.cljfmt.clj` files in directory hierarchy.

weavejester:
Yes, by design.

raycheung:
Oh, may I know the reason behind please?
Also, if so, when would the .cljfmt.edn be read?

I'm asking as I'm a vim-cljfmt user, but it couldn't have a way to integrate with respect to a custom config. The options computed in the cljfmt.main are private.

I would like to make it work as I love cljfmt but it doesn't integrate with my vim too well at the moment.

weavejester:
The `.cljfmt` files are designed as ways of configuring cljfmt when used as a standalone tool. When used via Leiningen it uses Leiningen's own configuration system (so `project.clj`, `profiles.clj`, etc.). Can you explain your use-case? I'm not sure why you'd want to use Leiningen to start cljfmt, but simultaneously not use Leiningen to configure cljfmt.

raycheung:
Thanks explaining.
The way current I'm using cljfmt is via vim, more precisely via vim-cljfmt. It does a quite indirect way to integrate.
More precisely, at https://github.com/venantius/vim-cljfmt/blob/master/plugin/cljfmt.vim#L41 , it builds a clojure string which is evaluated against the current editing file (buffer) in vim, then pop it back as the reformatted.

The issue is that if I want to configure any custom indents or so, there is no way the vim-cljfmt could easily work it out the final config (options) to pass to cljfmt.

It isn't a cljfmt issue per se. However it would be much nicer if there is a mechanism to trigger the reformat with all the hierarchical configured options working out.

weavejester:
Thanks for explaining. In this case, however, it looks like vim-cljfmt doesn't use the `.cljfmt.*` files for configuration?

raycheung:
You're right, it doesn't. I'm trying to make it works.
Could you advise which function call sequence would work?
Reading cljfmt.main, it appears to me the `.cljfmt.*` are read only in the `-main` with a series of private functions, and I couldn't obtain the result outside that function.

weavejester:
We'd likely have to write a public API for it, as functions like `load-configuration` are private, as you point out.

sirmspencer:
[...]

PR #302, opened 2023-04-13
Load cljfmt file with :load-dotfile? setting

sirmspencer:
Fixes issue #301 

Add option for lein cljfmt to read from the same .cljfmt.edn file as the cli version.

### DIFF
--- a/lein-cljfmt/src/leiningen/cljfmt.clj
+++ b/lein-cljfmt/src/leiningen/cljfmt.clj
@@ -21,6 +21,9 @@
 (defn- merge-default-options [options]
   (config/merge-configs config/default-config options))
 
+(defn- merge-file-options [{:keys [load-config-file?] :as options}]
+  (config/merge-configs (when load-config-file? (config/load-config)) options))
+
 (defn ^:no-project-needed cljfmt
   "Format Clojure source files."
   [project command & paths]
@@ -30,6 +33,7 @@
         options (-> (:cljfmt project)
                     (assoc :paths paths)
                     (assoc :project-root (:root project))
+                    (merge-file-options)
                     (merge-default-options))]
     (if leiningen.core.main/*info*
       (execute-command command options)


sirmspencer:
@weavejester Any comments on this PR?

sirmspencer:
> @weavejester Any comments on this PR?

@weavejester bump.  Your guidelines say to do so.

weavejester:
They certainly do. These past weeks have been a little busy for me, but I have some time now to review.

README.md
```diff
@@ -233,7 +237,7 @@ Indentation types are:
 * `:inner` -
   two character indentation applied to form arguments at a depth
   relative to a form symbol
-
```
weavejester, line 236: Let's remove this whitespace cleanup from the commit, so we have a clean diff.

cljfmt/src/cljfmt/main.clj
```diff
@@ -167,7 +167,7 @@
   (or (find-file-in-dir dir ".cljfmt.edn")
       (find-file-in-dir dir ".cljfmt.clj")))
 
-(defn- load-configuration []
+(defn load-configuration []
```
weavejester, line 170: Since this is a public function, let's give it a docstring. Something like:

```clojure
(defn load-configuration []
  "Load a `.cljfmt.edn` or `.cljfmt.clj` configuration file from the current
  directory or from the first parent directory to contain one."
  ...)
```

README.md
```diff
@@ -108,6 +108,10 @@ project:
 ```clojure
 :cljfmt {}
 ```
+#### Leiningen specific configuration
+* `:load-dotfile?` -
+  true if cljfmt read configuration from `.cljfmt.edn` or `.cljfmt.clj`.
+  Defaults to false.
```
weavejester, line 114: Let's expand on this a little:

```markdown
  true if the CLI configuration file (either `.cljfmt.edn` or `.cljfmt.clj`)
  should be merged with the configuration options in the Leiningen project
  file. The Leiningen options always take priority. Defaults to false.
```

weavejester (changes requested):
Some minor changes needed around documentation, otherwise good.

README.md
sirmspencer: Can it be a different commit then?

README.md
sirmspencer: > The Leiningen options always take priority

I don't think that is true as written.  Should I swap the merge order here 

```
(defn- merge-file-options [{:keys [load-dotfile?] :as options}]
  (main/merge-options options (when load-dotfile? (main/load-configuration))))

```

weavejester: Ah, yes, the Leiningen options should take priority, as that allows you the option to override the dotfile if you want.

README.md
weavejester: It should ideally be a separate PR, so we can look at the diff of this PR and see only the changes needed to implement it.

sirmspencer:
@weavejester I made the changes you requested.

README.md
```diff
@@ -109,6 +109,11 @@ project:
 :cljfmt {}
 ```
 
+#### Leiningen specific configuration
+
+* `:load-dotfile?` -
+  true if the CLI configuration file (either `.cljfmt.edn` or `.cljfmt.clj`) should be merged with the configuration options in the Leiningen project file. The Leiningen options always take priority. Defaults to false.
```
weavejester, line 115: Can you wrap this line at 72 characters, the same as the rest of the README?

sirmspencer:
> Can you wrap this line at 72 characters, the same as the rest of the README?
@weavejester Done.

sirmspencer:
@weavejester Bump

lein-cljfmt/src/leiningen/cljfmt.clj
```diff
@@ -30,7 +33,8 @@
         options (-> (:cljfmt project)
                     (dissoc :paths)
                     (assoc :project-root (:root project))
-                    (merge-default-options))]
+                    (merge-default-options)
+                    (merge-file-options))]
```
weavejester, line 37: These two lines are in the wrong order; the way it's set up right now, the default values will take priority over the dotfile. Instead, it should be: defaults -> dotfile -> leiningen. If you put `merge-default-options` after `merge-file-options` it should work (but test it please).

weavejester (changes requested):
Sorry for the delay - looks like there's a small bug in the code that needs to be fixed.

jakepearson:
Anything I can do to help get this over the line?

sirmspencer:
> Anything I can do to help get this over the line?

I'll look at it this weekend.

sirmspencer:
A bunch of changes went out between this PR and now.  There is a blocking bug.

sirmspencer:
@weavejester I was finally able to get this updated again.  Its tested and ready to go.

sirmspencer:
@weavejester Bump

weavejester:
Apologies for the delay. Since this PR was added, there's a non-dotfile config, `cljfmt.edn` and `cljfmt.clj`, so I think it probably makes more sense to use `:load-config-file?` as the option, unless you can think of a better name?

The README should also be updated accordingly, with the non-dotfile versions of the config. Also I think we can add a new section, `### Leiningen Options`, for Leiningen-specific options like this. It probably shouldn't be under formatting options, as it isn't anything directly to do with formatting.

weavejester:
Actually, if you exclude the README from the commit, I can probably add the README information myself later, which would probably be easiest!

sirmspencer:
@weavejester Update the key.  I will happily let you add to the readme instead.

weavejester:
Perfect, thanks!

PR #308, opened 2023-06-15
Fix load-config for edn files with :file-pattern

jean-lopes:
When trying to modify the :file-pattern via `.edn` config file we have one of the following scenarios:

**cljfmt.edn with regex pattern**

```edn
{:file-pattern #"very-weird-pattern"}
```
If we try to load this file, an exception is thrown:

> Execution error at cljfmt.config/read-config (config.clj:33).
No dispatch macro for: "

**cljfmt.edn with a string**

```edn
{:file-pattern #"very-weird-pattern"}
```
The config is loaded, but when trying to apply the regex it fails because it is a String and not a regex pattern.

> Execution error (ClassCastException) at cljfmt.tool/grep$fn (tool.clj:21).
class java.lang.String cannot be cast to class java.util.regex.Pattern (java.lang.String and java.util.regex.Pattern are in module java.base of loader 'bootstrap')

I run into this problem on v0.10.5


weavejester:
Yes, that's what I was thinking of. I need to consider if that's the best solution though, and to add in some configuration tests.

In the meantime, you can of course use `.cljfmt.clj` instead.

weavejester:
So I've had a think about this, and I think it probably is the best solution, and would maintain compatibility with clojure-lsp. We probably want to add some tests, but I believe the config namespace is lacking tests anyway, so I'll go ahead and add some in so there's something to build off.

PR #315, opened 2023-07-28
[FIX] Deep merge options for lein-cljfmt

### LINKED ISSUE #314: Options don't merge correctly (version 11)
merge-configs was replaced with merge, and now its not doing a deep merge on indents.  This commit https://github.com/weavejester/cljfmt/commit/e5eabdc14bbc4982d802a86aba51bae7e5c01ae9

I see it in lein-cljfmt. Defaults are being overridden by project.clj or cljfmt.edn. I suspect it affect all places in the code where merge-configs was replaced.

sirmspencer:
Fixes #314, but only for lein-cljfmt

### DIFF
--- a/cljfmt/src/cljfmt/tool.clj
+++ b/cljfmt/src/cljfmt/tool.clj
@@ -151,3 +151,17 @@
   [options]
   (let [opts (merge (config/load-config) options)]
     (fix-no-config opts)))
+
+(defn deep-merge-with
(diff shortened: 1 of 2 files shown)


sirmspencer:
To be honest this isn't 100% tested. I am getting `java.lang.NullPointerException: Cannot invoke "java.util.regex.Pattern.matcher(java.lang.CharSequence)" because "re" is null`

PR #331, opened 2024-01-02
feat: clean up after a successful install

raszi:
With this change we would remove the unnecessary tarball after a successful install. It is really useful in Docker build where we want to keep the image on a bare minimum.

clj-kondo has a similar step in their install scripts to clean up the leftovers:

https://github.com/clj-kondo/clj-kondo/blob/master/script/install-clj-kondo


velios:
@weavejester I was just passing by, but I wonder whether this PR will be accepted or rejected? looks small enough that a decision can be made quickly, it seems they simply forgot about it

PR #337, opened 2024-02-21
wip

juszczakn:


cljfmt/src/cljfmt/config.clj
```diff
@@ -59,6 +59,46 @@
 (defn- directory? [path]
   (some-> path io/file .getAbsoluteFile .isDirectory))
 
+(defn- -load-config
+  [path]
+  (->> (some-> path read-config)
+       (merge default-config)
+       (convert-legacy-keys)))
+
+(defn- deep-merge
```
juszczakn, line 68: just stole quick from medley as an example

cljfmt/src/cljfmt/config.clj
```diff
@@ -59,6 +59,46 @@
...
+                                  (deep-merge v v')
+                                  v')))
+                   (assoc m k v'))))]
+       (reduce merge-entry (or a {}) (seq b)))))
+  ([a b & more]
+   (reduce deep-merge (or a {}) (cons b more))))
+
+(defn- load-extra-configs
+  "Load any :extra-configs specified in the base config file.
+  Note: is not recursive."
```
juszczakn, line 93: I could just see recursive configs becoming a mess very quickly. If doing something smarter, like e.g. looking at the classpath, then maybe something like that could work? idk.

.gitignore
```diff
@@ -9,3 +9,5 @@ pom.xml.asc
 .nrepl-port
 reports
 .cpcache
+.clj-kondo
+.lsp
```
weavejester, line 13: Changes that aren't relevant to the PR should ideally be omitted.

cljfmt/src/cljfmt/config.clj
weavejester: Merging configurations isn't a deep-merge normally in cljfmt, so it shouldn't be for `:extra-configs`. Instead, it should be a normal merge.

README.md
```diff
@@ -284,6 +284,16 @@ In order to load the standard configuration file from Leiningen, add the
   Paths can also be passed as command line arguments. If the path is
   `-`, then the input is STDIN, and the output STDOUT.
 
+* `:extra-configs` -
+  additional config files that will also be imported and merged into
+  the base configuration. Is not recursive, so `:extra-configs` in
```
weavejester, line 289: It probably should be recursive.

cljfmt/src/cljfmt/config.clj
```diff
@@ -59,6 +59,48 @@
...
+                   (assoc m k v'))))]
+       (reduce merge-entry (or a {}) (seq b)))))
+  ([a b & more]
+   (reduce deep-merge (or a {}) (cons b more))))
+
+(defn- load-extra-configs
+  "Load any :extra-configs specified in the base config file.
+  Notes:
+    * Not recursive.
+    * Only select keys are merged (`:extra-indents`)."
```
weavejester, line 89: Docstrings for private functions should be omitted.

cljfmt/test_resources/cljfmt.edn
```diff
@@ -0,0 +1,4 @@
+{:extra-indents {com.foo/a [[:inner 0]]
+                 com.unrelated/a [[:inner 0]]}
+ :extra-configs ["extra1-cljfmt.edn"
+                 "extra2-cljfmt.edn"]}
```
weavejester, line 4: These resources should be put under a `cljfmt` folder, rather than at the root, to better group them and prevent any namespace clashes.

cljfmt/test/cljfmt/config_test.clj
```diff
@@ -1,10 +1,32 @@
...
          (config/convert-legacy-keys {:legacy/merge-indents? true
                                       :indents {'foo [[:inner 0]]}}))))
+
+(deftest test-load-config-extra-configs
+  (testing "`:extra-configs` allows you to import `:extra-indents`"
+    (is (= (config/load-config (io/resource "empty-cljfmt.edn"))
+           (-> (config/load-config (io/resource "cljfmt.edn"))
+               (dissoc :extra-configs)
+               (assoc :extra-indents {})))
+        "should only differ by `:extra-indents` (via `:extra-configs`). All other keys are ignored.")
```
weavejester, line 20: Lines should be within 80 characters, and output should be plaintext rather than markdown.

cljfmt/test/cljfmt/config_test.clj
```diff
@@ -1,10 +1,32 @@
 (ns cljfmt.config-test
   (:require [cljfmt.config :as config]
-            [clojure.test :refer [deftest is]]))
+            [clojure.test :refer [deftest is testing]]
+            [clojure.java.io :as io])
+  (:import (java.io File)))
```
weavejester, line 5: I believe in other files `(import [java.io File])` is used as the style.

cljfmt/src/cljfmt/config.clj
```diff
@@ -68,6 +110,4 @@
    (let [path (if (directory? path)
                 (find-config-file path)
                 path)]
-     (->> (some-> path read-config)
-          (merge default-config)
-          (convert-legacy-keys)))))
+     (load-extra-configs path))))
```
weavejester, line 113: Perhaps instead we separate out the merging of extra configs like so:

```clojure
(->> (some-> path read-config)
     (merge default-config)
     (merge-extra-configs)
     (convert-legacy-keys))
```

weavejester (changes requested):
Thanks for the update, I've reviewed it and have some suggestions.

.gitignore
juszczakn: These are both standard things that clojure-lsp/IDE's might pull in, so figured they might be useful to add. I can yank it though if you really don't want them included

cljfmt/src/cljfmt/config.clj
juszczakn: Well, I don't really think it would make sense personally for each config to be wiping each others :extra-indents, as the whole point in my mind would be for them to be merged together. I assume you mean that it would make sense to just `merge` the :extra-indents into one another? I can do that pretty easily if that works?

cljfmt/src/cljfmt/config.clj
juszczakn: Couple of ways I could interpret this feedback:

* Either that's an argument that the code is mistructured and this is a code smell that I thought it was worthy of a comment, and should be refactored (self-documenting, or extracted from ns, or some other thing?)
* docstrings/comments aren't helpful or are confusing in private fn's? or perhaps is messes with some sort of automation or something?

I'm unsure how to accept the latter. I can just yank docstrings from private fn's I guess? But this feedback feels like shorthand for the former I'd assume?

cljfmt/test_resources/cljfmt.edn
juszczakn: sounds good, can do

cljfmt/test/cljfmt/config_test.clj
juszczakn: leftover import, will just yank

cljfmt/src/cljfmt/config.clj
juszczakn: yea I can restructure to do it that way

README.md
juszczakn: Yea I could see that; I was worried about how this might be confusing or hard to debug. Some odd corner cases I was thinking about:

* You pull in a config from a lib A, which references lib B, but you never pull in lib B's config, they end up missing or causing errors?
* Could ordering get messy? thinking of multiple libraries using some other base library
* Could the relative path references get messed up, depending on how they're pulled in?
* Probably some others I'm missing? 

If it's not recursive though, I could see inconveniences there as well. Not sure.

cljfmt/src/cljfmt/config.clj
juszczakn: done

cljfmt/test/cljfmt/config_test.clj
juszczakn: fixed, think I got them all?

cljfmt/test_resources/cljfmt.edn
juszczakn: I put them under `cljfmt/config_test`, to further ns them. Can change it to just cljfmt if that's not acceptable

cljfmt/src/cljfmt/config.clj
juszczakn: I just removed the docstr w/ the rewrite to change it to be `merge-extra-configs`, let me know if there's some other thing you want done

cljfmt/src/cljfmt/config.clj
juszczakn: I've updated it to merge the :extra-indents maps together, and then merge the final :extra-indents back into the base config. If that's not what you meant or Ok let me know

.gitignore
weavejester: They are useful, but they're just not relevant to the current PR. A PR should *only* contain code to implement the feature/fix described.

cljfmt/src/cljfmt/config.clj
weavejester: Hm, that's a fair point. Does `deps.edn` solve this in any particular way?

I don't like the idea of recursively merging configs, as that makes the process purely additive, and makes it different from how the default configuration is handled, and how options are handled.

I'm beginning to think that this issue might be better solved via a different means entirely. For example, if the use-case is just supporting extra indentations in libraries, we could create a indent resource in a standard location.

juszczakn: > Hm, that's a fair point. Does deps.edn solve this in any particular way?

I'm not familiar with deps.edn at all really, so unsure

> I don't like the idea of recursively merging configs, as that makes the process purely additive, and makes it different from how the default configuration is handled, and how options are handled.

In my mind, this is new behavior is somewhat consistent with how :extra-indents works today, adding to the default :indents . I was also thinking that, since the base config has the final say, it could overwrite any values pulled in from downstream that it doesn't agree with, but that may be less than ideal, not sure.

> I'm beginning to think that this issue might be better solved via a different means entirely. For example, if the use-case is just supporting extra indentations in libraries, we could create a indent resource in a standard location.

I'm not sure what you mean by this, but I'm open to other ideas/solutions

.gitignore
juszczakn: Ok I can yoink them then if them's the rules.

cljfmt/src/cljfmt/config.clj
weavejester: > I'm not sure what you mean by this, but I'm open to other ideas/solutions

clj-kondo has a mechanism for [importing configurations from libraries](https://github.com/clj-kondo/clj-kondo/blob/master/doc/config.md#exporting-and-importing-configuration). We could do something similar with cljfmt.

The way it would work is that cljfmt can optionally be given a classpath, which it will then search through to find indentation files in some predictable location. For example `cljfmt.exports/*/*/indent.edn`.

PR #342, opened 2024-03-15
Feature: reporting fn to allow more ergonomic use as a library

cap10morgan:
This is a follow-up to #333 with the implementation outlined by @weavejester over there.

I think the main decision I made in here was the name and signature of the "library entry point" fns. I went with `check-paths` and `fix-paths` that take a paths collection as a first arg and then an optional options map. I did this b/c the primary purpose of this refactor was to create library-friendly entry points that check / fix paths like the CLI tool entry points do but w/o doing things like printing to stdout or exiting the whole process.

But I don't have strong opinions about any of that, so let me know if you'd like to see any changes there or anywhere else.

### DIFF
--- a/cljfmt/src/cljfmt/tool.clj
+++ b/cljfmt/src/cljfmt/tool.clj
@@ -3,19 +3,12 @@
             [cljfmt.core :as cljfmt]
             [cljfmt.diff :as diff]
             [cljfmt.io :as io]
-            [clojure.java.io :as cio]
-            [clojure.stacktrace :as st]))
+            [cljfmt.report :as report]))
 
-(def ^:dynamic *no-output* false)
 (def ^:dynamic *verbose* false)
 
-(defn- warn [& args]
-  (when-not *no-output*
-    (binding [*out* *err*]
-      (apply println args))))
-
 (defn- trace [& args]
-  (when *verbose* (apply warn args)))
+  (when *verbose* (apply report/warn args)))
 
 (defn- grep [re dir]
   (filter #(re-find re (io/relative-path % dir)) (io/list-files dir)))
@@ -30,26 +23,21 @@
 (defn- reformat-string [options s]
   ((cljfmt/wrap-normalize-newlines #(cljfmt/reformat-string % options)) s))
 
-(defn- project-path [{:keys [project-root]} file]
-  (->> (or project-root ".") cio/file (io/relative-path file)))
-
 (defn- format-diff
   ([options file]
    (let [original (io/read-file file)]
      (format-diff options file original (reformat-string options original))))
   ([options file original revised]
-   (let [filename (project-path options file)
+   (let [filename (io/project-path options file)
          diff     (diff/unified-diff filename original revised)]
      (if (:ansi? options)
        (diff/colorize-diff diff)
        diff))))
 
-(def ^:private zero-counts {:okay 0, :incorrect 0, :error 0})
-
 (defn- check-one [options file]
-  (trace "Processing file:" (project-path options file))
+  (trace "Processing file:" (io/project-path options file))
   (let [original (io/read-file file)
-        status   {:counts zero-counts :file file}]
+        status   {:counts report/zero-counts :file file}]
     (try
       (let [revised (reformat-string options original)]
         (if (not= original revised)
@@ -62,56 +50,17 @@
             (assoc-in [:counts :error] 1)
             (assoc :exception ex))))))
 
-(defn- print-stack-trace [ex]
-  (when-not *no-output*
-    (binding [*out* *err*]
-      (st/print-stack-trace ex))))
-
-(defn- print-file-status [options status]
-  (let [path (project-path options (:file status))]
-    (when-let [ex (:exception status)]
-      (warn "Failed to format file:" path)
-      (print-stack-trace ex))
-    (when (:reformatted status)
-      (warn "Reformatted" path))
-    (when-let [diff (:diff status)]
-      (warn path "has incorrect formatting")
-      (println diff))))
-
-(defn- exit [counts]
-  (when-not (zero? (:error counts 0))
-    (System/exit 2))
-  (when-not (zero? (:incorrect counts 0))
-    (System/exit 1)))
-
-(defn- print-final-count [counts]
-  (let [error     (:error counts 0)
-        incorrect (:incorrect counts 0)]
-    (when-not (zero? error)
-      (warn error "file(s) could not be parsed for formatting"))
-    (when-not (zero? incorrect)
-      (warn incorrect "file(s) formatted incorrectly"))
-    (when (and (zero? incorrect) (zero? error))
-      (warn "All source files formatted correctly"))))
-
-(defn- merge-counts
-  ([]    zero-counts)
-  ([a]   a)
-  ([a b] (merge-with + a b)))
-
 (defn check-no-config
   "The same as `check`, but ignores dotfile configuration."
-  [options]
-  (let [map*   (if (:parallel? options) pmap map)
-        counts (->> (:paths options)
-                    (mapcat (partial find-files options))
-                    (map* (partial check-one options))
-                    (map (fn [status]
-                           (print-file-status options status)
-                           (:counts status)))
-                    (reduce merge-counts))]
-    (print-final-count counts)
-    (exit counts)))
+  [{:keys [parallel? paths report] :as options
+    :or   {report report/console}}]
+  (let [context (report :check/initial options nil)
+        map*    (if parallel? pmap map)
+        summary (->> paths
+                     (mapcat (partial find-files context))
+                     (map* (partial check-one context))
+                     (reduce #(report :check/file %1 %2) context))]
+    (report :check/summary summary nil)))
 
 (defn check
   "Checks that the Clojure paths specified by the :paths option are
@@ -121,7 +70,7 @@
     (check-no-config opts)))
 
 (defn- fix-one [options file]
-  (trace "Processing file:" (project-path options file))
+  (trace "Processing file:" (io/project-path options file))
   (let [original (io/read-file file)]
     (try
       (let [revised  (reformat-string options original)
@@ -138,13 +87,15 @@
 
 (defn fix-no-config
   "The same as `fix`, but ignores dotfile configuration."
-  [options]
-  (let [files (recursively-find-files options)
-        map*  (if (:parallel? options) pmap map)]
-    (->> files
-         (map* (partial fix-one options))
-         (map (partial print-file-status options))
-         dorun)))
+  [{:keys [parallel? report] :as options
+    :or   {report report/console}}]
+  (let [context (report :fix/initial options nil)
+        files   (recursively-find-files context)
+        map*    (if parallel? pmap map)
+        summary (->> files
+                     (map* (partial fix-one context))
+                     (reduce #(report :fix/file %1 %2) context))]
+    (report :fix/summary summary nil)))
 
 (defn fix
   "Fixes the formatting for all files specified by the :paths option."
--- a/README.md
+++ b/README.md
@@ -134,7 +134,8 @@ project file. See the [configuration][] section for more details.
 
 ### Library
 
-cljfmt can be run as a library that formats a string of Clojure code.
+cljfmt can be run as a library that formats a string of Clojure code or
+recursively checks / fixes paths like the CLI tool.
 First, add the dependency:
 
 ```edn
@@ -143,6 +144,8 @@ First, add the dependency:
 
 Then use the library:
 
+#### Checking strings of code:
+
 ```clojure
 (require '[cljfmt.core :as fmt])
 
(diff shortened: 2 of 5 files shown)


weavejester:
Thanks for the PR, and my apologies for not being clear about the design I was suggesting. In clojure.test and other similar libraries, the reporting function reacts to events like `:pass`, `:fail` and `:summary`, and then decides how it will output these. We could do something similar with cljfmt. Consider a multimethod:

```clojure
(defmulti cli-report
  (fn [event context data] event))
```

Used like:

```clojure
(defn check-no-config [{:keys [parallel? paths report] :as options}]
  (let [map*    (if parallel? pmap map)
        summary (->> paths
                     (mapcat (partial find-files options))
                     (map* (partial check-one options))
                     (reduce #(report :check/file %1 %2) options))]
    (report :check/summary summary nil)))
```

The reporter would be responsible for reporting data for the user. This includes both outputting on a file-by-file basis, and collating the data in a context map for use in the summary. We could use the options as the initial context, and add a `::count` key to it.

I'd imagine we'd want at least 4 events: `:check/file`, `:check/summary`, `:fix/file` and `:fix/summary`. Maybe a `:check/initial` and `:fix/initial` as well that would be no-op by default, but available for reporters that want to setup something like a progress bar.

cap10morgan:
@weavejester Makes sense. I'll work on moving it in that direction next chance I get. Thanks for the feedback!

cap10morgan:
@weavejester Any other TODO's here? Happy to refine if this needs some more work before merging.

cap10morgan:
@weavejester just checking in again to see where we're at with this PR

weavejester:
Apologies for the delay in looking at this. I'll review the code now.

cljfmt/src/cljfmt/tool.clj
```diff
@@ -99,25 +100,77 @@
...
+
+      (and file diff) (assoc-in [:incorrect path] diff)
+      (and file exception) (assoc-in [:error path] exception))))
+
+(defmulti report-cli (fn [event & _] event))
+(defmulti report-lib (fn [event & _] event))
+
+(defmethod report-cli :check/initial
+  [_ _ _]
+  nil)
```
weavejester, line 123: Can you write these methods a little more concisely? e.g.

```clojure
(defmethod report-cli :check/initial [_ _ _])
```

And:

```clojure
(defmethod report-lib :fix/file [_ _ data] data)
```

cljfmt/src/cljfmt/tool.clj
```diff
@@ -99,25 +100,77 @@
...
+      true (update :counts #(merge-with + % (:counts b)))
+
+      (and file (= 1 (-> b :counts :incorrect)))
+      (assoc-in [:incorrect path] "has incorrect formatting")
+
+      (and file diff) (assoc-in [:incorrect path] diff)
+      (and file exception) (assoc-in [:error path] exception))))
+
+(defmulti report-cli (fn [event & _] event))
+(defmulti report-lib (fn [event & _] event))
```
weavejester, line 119: These multimethods should explicitly define their signatures, e.g.

```clojure
(defmulti report-cli (fn [event _context _data] event))
```

Can you also group the `report-cli` and `report-lib` multimethods together?

cljfmt/src/cljfmt/tool.clj
```diff
@@ -99,25 +100,77 @@
...
-  (let [map*   (if (:parallel? options) pmap map)
-        counts (->> (:paths options)
-                    (mapcat (partial find-files options))
-                    (map* (partial check-one options))
-                    (map (fn [status]
-                           (print-file-status options status)
-                           (:counts status)))
-                    (reduce merge-counts))]
+(defn- merge-statuses
+  [a b]
```
weavejester, line 104: Prefer

```clojure
(defn- merge-statuses [a b]
```

There's no need to add in extra whitespace when it doesn't aid readability.

cljfmt/src/cljfmt/tool.clj
```diff
@@ -99,25 +100,77 @@
...
+                    (config/load-config) options)]
+    (check-no-config opts)))
+
+(defn check-paths
+  "Runs a check on the provided paths and returns a map detailing the results."
+  [paths & [options]]
+  (let [opts (merge {:report report-lib}
+                    (config/load-config) options
+                    {:paths paths})]
     (check-no-config opts)))
```
weavejester, line 174: Why add this function? Why not make it equivalent to using `report-lib` with `check`?

cljfmt/src/cljfmt/tool.clj
```diff
@@ -4,7 +4,8 @@
             [cljfmt.diff :as diff]
             [cljfmt.io :as io]
             [clojure.java.io :as cio]
-            [clojure.stacktrace :as st]))
+            [clojure.stacktrace :as st])
+  (:import (java.io File)))
```
weavejester, line 8: Prefer `(:import [java.io File])`

cljfmt/src/cljfmt/tool.clj
```diff
@@ -99,25 +100,77 @@
...
-  [options]
-  (let [map*   (if (:parallel? options) pmap map)
-        counts (->> (:paths options)
-                    (mapcat (partial find-files options))
-                    (map* (partial check-one options))
-                    (map (fn [status]
-                           (print-file-status options status)
-                           (:counts status)))
-                    (reduce merge-counts))]
+(defn- merge-statuses
```
weavejester, line 103: I don't understand the purpose of `merge-statuses`. Can you explain?

cljfmt/src/cljfmt/tool.clj
```diff
@@ -99,25 +100,77 @@
...
+(defn check-no-config
+  "The same as `check`, but ignores dotfile configuration."
+  [{:keys [parallel? paths report] :as options
+    :or   {report report-cli}}]
+  (report :check/initial options nil)
+  (let [map*    (if parallel? pmap map)
+        summary (->> paths
+                     (mapcat (partial find-files options))
+                     (map* (partial check-one options))
+                     (map #(report :check/file options %)))]
```
weavejester, line 157: The `report` should be a `reduce`, not a map. The reporting functions should be a type signature of:

```clojure
(report event context data) -> new-context
```

That way the type signature of the multi-method remains consistent. See the example in my earlier comment.

weavejester (changes requested):
Thanks for the commits. I have some questions/suggestions. The primary issue to be solved is that the reporting function needs a consistent type signature.

cljfmt/src/cljfmt/tool.clj
cap10morgan: Well, the entire point of this PR is to make this more ergonomic to use as a library. So this is the library entry point fn for checking paths.

Not sure I understand this:
> Why not make it equivalent to using report-lib with check?

cljfmt/src/cljfmt/tool.clj
cap10morgan: It's the counterpart to `merge-counts` when you want a fuller report of what was done, which you often do when using this as a library.

cljfmt/src/cljfmt/tool.clj
weavejester: I mean something like:

```clojure
(check {:paths ["src"], :report report-lib})
```

cljfmt/src/cljfmt/tool.clj
cap10morgan: Yeah, I started with that. It's a little more complex and messy that way. Instead of just returning what is inherently a list of results, you have to stuff it into the original context somewhere, which starts out as the `options` map. So then the final return value is all of those options + wherever you stuffed the results.

I have a working refactor that puts them into a `:results` key. Happy to push that if you still prefer it.

weavejester:
To give a little more context, this was the design I had in mind for the CLI reporter. Note that the method signature remains consistent: `event context data -> context`, with the initial context being the options map.

```clojure
(defmulti report-cli
  (fn [event _context _data] event))

(defmethod report-cli :check/init [_ context _] context)

(defmethod report-cli :fix/init [_ context _] context)

(defmethod report-cli :check/file [_ context file-status]
  (print-file-status context file-status)
  (update context :counts (fnil merge-counts {}) (:counts file-status)))

(defmethod report-cli :fix/file [_ context file-status]
  (print-file-status context file-status)
  context)

(defmethod report-cli :check/summary [_ {:keys [counts]} _]
  (print-final-count counts)
  (exit counts))

(defmethod report-cli :fix/summary [_ _ _])
```

cap10morgan:
OK @weavejester I just pushed some changes that I think address your feedback

cljfmt/src/cljfmt/tool.clj
cap10morgan: I will also note that this is a big part of why I started with a different namespace. Naming this fn in here is tricky in terms of revealing its intention (and so was naming the other namespace as you probably recall; so that's perhaps not a great solution either).

But the idea is you would call this fn when using cljfmt as a library to check paths (instead of strings), and the existing fn(s) are still for internal CLI usage. I'm open to ideas about how to better name this, document that intention, etc.

But I do think it makes sense to have a "library entrypoint" fn that doesn't require setting options you'll want every time to make it be a well-behaved library (e.g. not calling `System/exit`, not printing to stdout, etc.).

cljfmt/src/cljfmt/tool.clj
```diff
@@ -136,18 +180,45 @@
...
-         dorun)))
+  [{:keys [parallel? report] :as options
+    :or   {report report-cli}}]
+  (let [context (report :fix/initial options nil)
+        files   (recursively-find-files context)
+        map*    (if parallel? pmap map)
+        summary (->> files
+                     (map* (partial fix-one context))
+                     (reduce #(report :fix/file %1 %2) context)
+                     doall)]
```
weavejester, line 208: `doall` isn't required if there is a `reduce` before it.

cap10morgan: Ah, yep. Meant to revisit that. Thanks!

cljfmt/src/cljfmt/tool.clj
weavejester: I think I'd rather omit the `check-paths` and `fix-paths` wrapper functions for now. I don't think it's a huge deal to add an extra `:report` argument to cater for a relatively niche use-case.

I think it would also be a good idea to move `report-cli` to `cljfmt.report/console` and `report-lib` to `cljfmt.report/clojure`. However, don't feel you need to make that chage; I can always move things around after the PR is merged.

cap10morgan: OK, I can make both changes

cljfmt/src/cljfmt/tool.clj
```diff
@@ -99,25 +100,68 @@
...
+        diff       (:diff b)
+        exception  (:exception b)]
+    (cond-> a
+            true (update :counts #(merge-with + % (:counts b)))
+
+            (and file (= 1 (-> b :counts :incorrect)))
+            (assoc-in [:incorrect path] "has incorrect formatting")
+
+            (and file diff) (assoc-in [:incorrect path] diff)
+            (and file exception) (assoc-in [:error path] exception))))
```
weavejester, line 115: Maybe something more like:

```clojure
(defn- merge-statuses [results {:keys [file diff exception counts]}]
  (let [path (some-> file .getPath)]
    (-> results
        (update :counts merge-counts counts)
        (cond-> (and path (> (:incorrect counts) 1))
          (assoc-in [:incorrect-files path] {}))
        (cond-> (and path diff)
          (assoc-in [:incorrect-files path :diff] diff))
        (cond-> (and path exception)
          (assoc-in [:errored-files path :exception] exception)))))
```

cljfmt/src/cljfmt/tool.clj
cap10morgan: Would you prefer these changes as new, separate commits? Or should I just continue to amend the existing HEAD commit on this branch?

cap10morgan: I'll go ahead and push separate commits for easier review. Happy to squash later if desired.

cap10morgan: When I moved the reporting fns to the new `cljfmt.report` ns, it required one other change to avoid a circular dependency: The `project-path` fn is used by both `cljfmt.tool` fns and `cljfmt.report` fns, so it needed to be in a namespace they both required. `cljfmt.io` seemed like a natural fit, so I moved it there.

cap10morgan:
@weavejester Should be ready for review again

cljfmt/src/cljfmt/tool.clj
```diff
@@ -62,66 +50,34 @@
...
+                     (map* (partial check-one context))
+                     (reduce #(report :check/file %1 %2) context))]
+    (report :check/summary summary nil)))
 
 (defn check
   "Checks that the Clojure paths specified by the :paths option are
   correctly formatted."
   [options]
-  (let [opts (merge (config/load-config) options)]
+  (let [opts (merge {:report report/console}
```
weavejester, line 75: Is the merge of `{:report report/console}` necessary given the `:or` set in `check-no-config`?

And similarly for `fix` and `fix-no-config`?

weavejester (commented):
Thanks for all your work! I just have one query this time.

cljfmt/src/cljfmt/tool.clj
cap10morgan: Good catch. No they are not! 😁 Will fix shortly

cap10morgan: Fixed in https://github.com/weavejester/cljfmt/pull/342/commits/ddf28c0b753ebd0fd51015f800c59b3133ec0972

weavejester:
Thanks for your work on this. It all looks good. Can you squash your commits down and give it an appropriate commit message?

cap10morgan:
> Thanks for your work on this. It all looks good. Can you squash your commits down and give it an appropriate commit message?

Done

weavejester:

```
Add :report option to check and fix

Allows for a custom reporter function when checking/fixing files.
```

I think that's perhaps a little clearer.

cap10morgan:
> Thanks! Could you change the commit message to:
> 
> ```
> Add :report option to check and fix
> 
> Allows for a custom reporter function when checking/fixing files.
> ```
> 
> I think that's perhaps a little clearer.

done

cap10morgan:
@weavejester https://github.com/weavejester/cljfmt/pull/342/commits/69ca1078a91d5f6222b5feca3d32642a7f21d7bf fixes the CI issue. It was flagging that I hadn't wired up the `tool/exit` fn into the new event-based workflow. It should now mimic previous behavior. Let me know if that looks good and I'll re-squash.

weavejester:
Thanks for the fix, and sorry for the delay again. It's been a busy week! It all looks good if you want to squash the changes.

cap10morgan:
@weavejester done

cap10morgan:
@weavejester Anything else needed here before merging?

weavejester:
Sorry for the delay! I've been away for a couple of weeks. I don't think there's anything left to do, so I'll merge once the checks are done, but I may delay the release a little as I think about the format of the reporters.

PR #344, opened 2024-05-26
Fix #45: Thread first changes the indentation of the inner forms

### LINKED ISSUE #45: Threaded cond-> indents strangely
`cond->` and `cond->>` fit neatly into bigger `->` forms, but the default indenting rules don't seem to work for that usage when conditions and threaded forms are each on their own line.

So cljfmt likes this:
```clojure
(-> v
    (cond->
     a b
     c d))
```
But wants to change this:
```clojure
(-> v
    (cond->
      a
      b
      c
      d))
```
like so:
```diff
 (-> v
     (cond->
-      a
+     a
       b
       c
       d))
```
It wants the first arg indented 1 space, the rest indented 2.

(Incidentally my out of date version of cider doesn't like either of those indentations and has its own bad ideas about always treating the first arg specially.)

lambdank:
Here is an attempt to solve issue #45. 

The concrete issue is nesting a `->` with a `cond->` inside. Because `cond->` has indent `[[:block 1]]`, the parameters will not be indented in pairs as intended.
As mentioned in this [comment](https://github.com/weavejester/cljfmt/issues/45#issuecomment-122682926
), the indentation of an inner form within a `->` does not take the threading into account.

The proposed solution is to take the parent node into account, when determining the index of the indentation of the current node. 
The check is: 
- Is the parent form a thread-first macro? 
- Apart from this, there is the special case, which is the first argument to a thread-first, which should be indented as usual.

I just made this fix for thread-first macros in Clojure core. We could maybe do a regex instead? 

Please let me know, if I need to do any documentation or something else is missing.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -280,10 +280,22 @@
     (or (symbol-matches-key? (fully-qualified-symbol possible-sym context) key)
         (symbol-matches-key? (remove-namespace possible-sym) key))))
 
+(defn- parent-is-thread-first? [zloc context]
+  (some #(form-matches-key? zloc % context) #{'-> 'as-> 'some-> 'cond->}))
+
+(defn- is-first-parameter? [zloc]
+  (= (z/right (z/leftmost zloc)) zloc))
+
+(defn- adjust-idx [zloc idx context]
+  (cond-> idx
+    (and (parent-is-thread-first? zloc context) (not (is-first-parameter? zloc)))
+    (some-> dec (max 0))))
+
 (defn- inner-indent [zloc key depth idx context]
-  (let [top (nth (iterate z/up zloc) depth)]
+  (let [top (nth (iterate z/up zloc) depth)
+        adjusted-idx (adjust-idx (z/up zloc) idx context)]
     (when (and (form-matches-key? top key context)
-               (or (nil? idx) (index-matches-top-argument? zloc depth idx)))
+               (or (nil? idx) (index-matches-top-argument? zloc depth adjusted-idx)))
       (let [zup (z/up zloc)]
         (+ (margin zup) (indent-width zup))))))
 
@@ -302,9 +314,10 @@
 
 (defn- block-indent [zloc key idx context]
   (when (form-matches-key? zloc key context)
-    (let [zloc-after-idx (some-> zloc (nth-form (inc idx)))]
+    (let [adjusted-idx (adjust-idx (z/up zloc) idx context)
+          zloc-after-idx (some-> zloc (nth-form (inc adjusted-idx)))]
       (if (and (or (nil? zloc-after-idx) (first-form-in-line? zloc-after-idx))
-               (> (index-of zloc) idx))
+               (> (index-of zloc) adjusted-idx))
         (inner-indent zloc key 0 nil context)
         (list-indent zloc context)))))
 
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -692,7 +692,39 @@
          ["#:clj {:a :b"
           ":c :d}"]
          ["#:clj {:a :b"
-          "       :c :d}"]))))
+          "       :c :d}"]))) 
+
+  (testing "thread first"
+    (is (reformats-to?
+          ["(-> v"
+           "(cond->"
+           "a b"
+           "c d))"]
+          ["(-> v"
+           "    (cond->"
+           "      a b"
+           "      c d))"]))
+    (is (reformats-to?
+          ["(cond-> v"
+           "a b"
+           "c d)"]
+          ["(cond-> v"
+           "  a b"
+           "  c d)"]))
+    (is (reformats-to?
+          ["(-> v"
+           "(cond-> a b"
+           "c d))"]
+          ["(-> v"
+           "    (cond-> a b"
+           "            c d))"]))
+    (is (reformats-to?
+          ["(-> (cond-> a"
+           "odd? inc)"
+           "inc)"]
+          ["(-> (cond-> a"
+           "      odd? inc)"
+           "    inc)"]))))
 
 (deftest test-remove-multiple-non-indenting-spaces
   (let [opts {:remove-multiple-non-indenting-spaces? true}]


cljfmt/src/cljfmt/core.cljc
```diff
@@ -280,10 +280,22 @@
     (or (symbol-matches-key? (fully-qualified-symbol possible-sym context) key)
         (symbol-matches-key? (remove-namespace possible-sym) key))))
 
+(defn- parent-is-thread-first? [zloc context]
+  (some #(form-matches-key? zloc % context) #{'-> 'as-> 'some-> 'cond->}))
```
weavejester, line 284: A better name for this would be: `form-matches-thread-macro?`. The list of threading symbols should also be configurable based on the context, and I think this might need to be refactored so we only call `form-symbol` once.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -280,10 +280,22 @@
...
+
+(defn- is-first-parameter? [zloc]
+  (= (z/right (z/leftmost zloc)) zloc))
```
weavejester, line 287: In Clojure, by convention we omit the `is` from predicates, and use "argument" over "parameter". So, `first-argument?` would be a more idiomatic name.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -280,10 +280,22 @@
...
+
+(defn- adjust-idx [zloc idx context]
+  (cond-> idx
+    (and (parent-is-thread-first? zloc context) (not (is-first-parameter? zloc)))
+    (some-> dec (max 0))))
```
weavejester, line 292: Instead of `adjust-idx` which doesn't convey *why* the index is being adjusted, perhaps instead we just split out the predicate:

```clojure
(defn- in-thread-macro? [zloc context]
  (and (form-matches-thread-macro? zloc) (not (first-argument? zloc))))
```

Then we could write:

```clojure
(cond-> idx (in-thread-macro? zloc context) dec)
```

I'm unsure why you have `(some-> dec (max 0))` instead of just `dec`. Under what circumstances would the index be less than or equal to zero?

weavejester (changes requested):
Thanks for the PR. This obviously needs to be configurable before it can be merged. Ideally this would happen through an indentation rule, e.g. `:thread` to complement `:block`, though that might be a little tricky.

I think I need to take another look at the indentation logic anyway. I don't have a lot of time over the next couple of weeks, but I'll see if I can squeeze in a few hours to go over the indentation system.

cljfmt/src/cljfmt/core.cljc
lambdank: Wouldn't `idx` be zero in this case, if I configured `{cond-> [[:block 0]]}`?
There is a check for `(nil? idx)` in `inner-indent` already. Isn't it nil when there is no indentation rule, like `(-> x foo bar)`?

lambdank:
> Thanks for the PR. This obviously needs to be configurable before it can be merged. Ideally this would happen through an indentation rule, e.g. `:thread` to complement `:block`, though that might be a little tricky.
> 
> I think I need to take another look at the indentation logic anyway. I don't have a lot of time over the next couple of weeks, but I'll see if I can squeeze in a few hours to go over the indentation system.

As I see it, the tricky part about having a `:thread` indentation rule is, that it applies to the children and not to the node itself.
I could do something where I create a `:thread` rule, just indent them with the default indentation and gather them in the context to a use them as a replacement for the hardcoded `#{'-> 'as-> 'some-> 'cond->}`?

cljfmt/src/cljfmt/core.cljc
weavejester: The index is the position of the child in the parent form, IIRC. So, come to thing about it, we could change `(not (first-argument? zloc))` to `(> idx 1)`.

lambdank: Doesn't the index come from the :indents as specified [here](https://github.com/weavejester/cljfmt/blob/master/docs/INDENTS.md)?
So for `:block` it uses default indenting until we reach the index. So if we are in a thread, we should basically hit that index one spot before (ie. decrease the index). If the index is 0, we should keep it at 0.
For `:inner` when index is set, it is restricted to that particular index. So that should also occur one spot earlier (ie. decrease the index). If the index is 0, then we should avoid using this indentation rule, since the spot it was meant for is not present.

I also oversaw the `depth` argument of inner. Here we should check whether the parent of the `top` zloc is wrapped in a thread. I'll commit a fix.

Hope it makes sense.

weavejester: Ah, yes, I believe you're right. I need to refresh myself on the indentation logic as it's been a while since I've looked at it.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -280,10 +280,25 @@
     (or (symbol-matches-key? (fully-qualified-symbol possible-sym context) key)
         (symbol-matches-key? (remove-namespace possible-sym) key))))
 
+(defn- form-matches-thread-macro? [zloc context]
+  (let [possible-sym (form-symbol zloc)
+        fully-qualified-sym (fully-qualified-symbol possible-sym context)
+        sym-without-namespace (remove-namespace possible-sym)]
+    (some #(or (symbol-matches-key? fully-qualified-sym %)
+               (symbol-matches-key? sym-without-namespace %))
+          #{'-> 'as-> 'some-> 'cond->})))
```
weavejester, line 289: What about doing it this way:

```clojure
(defn- form-symbol-matches? [zloc pred context]
  (let [possible-sym (form-symbol zloc)]
    (or (pred (fully-qualified-symbol possible-sym context))
        (pred (remove-namespace possible-sym)))))

(defn- form-matches-key? [zloc key context]
  (form-symbol-matches? zloc #(symbol-matches-key? % key) context))

(defn- form-matches-thread-macro? [zloc context]
  (form-symbol-matches? zloc #{'-> 'as-> 'some-> 'cond->} context))

**Comment by weavejester** [comment]
> As I see it, the tricky part about having a `:thread` indentation rule is, that it applies to the children and not to the node itself. I could do something where I create a `:thread` rule, just indent them with the default indentation and gather them in the context to a use them as a replacement for the hardcoded `#{'-> 'as-> 'some-> 'cond->}`?

Yes, from a user perspective, we'd expect indentation rules to be all in one place, but threaded indentation is tricky because it depends on both the indentation rules for the current form and the indentation rules for the outer form.

It gets even trickier when we consider a form like `cond->`, where only the *even* arguments are threaded. I'm unsure of a good way to express that as a rule, unless we add in meta-indexes like `:even` and `:odd` to complement integer indexes.

**Comment by lambdank** [comment]
All right, I thought a bit about it and gave it another shot.
Just to sum up the changes in the last commit:
- I introduced a new rule called `:thread`. As can be seen in the `indenter-fn` defmethod, it does not directly influence the form itself, but rather influences the child forms. Therefore, I gather them in the context. This rule can have take an `idx`, which correspondes to at which argument the threading starts (eg. for `as->` it's 3). This can also be `:even` or `:odd`, where it maches every other form. Here the idx is hardcoded to 2 (which might not be necessary, coming to think of it). 
- Most of the complexity lies in `in-thread-macro?`. Before we should just look at the node and it's parent, but since we now allow even/odd, we also need to check, whether the parent is in a threading macro. If it is, we raise the position by one, such that it reflects the "actual" position. This means that this is a recursive function, which stops when the parent is the root or not in a threading macro.

I haven't written any documentation yet, since I wanted to hear your take on the approach @weavejester, before doing that work. Let me now what you think.

**Comment by lambdank** [comment]
I hope this test:
`(is (reformats-to?
         ["(cond-> a"
          "(cond-> 1
                  odd? inc) 
          (cond-> a b 
                  c d))"]
         ["(cond-> a"
          "  (cond-> 1"
          "    odd? inc)"
          "  (cond-> a b"
          "          c d))"]))`
might reflect the recursion.

File: cljfmt/src/cljfmt/core.cljc [hunk 5]
```diff
@@ -275,15 +275,41 @@
...
+(defn- form-matches-thread-macro? [zloc context]
+  (form-symbol-matches? zloc (:threads context) context))
+
+(defn- get-zipper-position [zloc]
+  (loop [idx 0
+         loop-zloc (z/leftmost zloc)]
+    (cond
+      (= zloc loop-zloc) idx
+      (= loop-zloc (z/rightmost zloc)) nil
+      :else (recur (inc idx) (z/right loop-zloc)))))
```
weavejester, line 295: How does this differ from `index-of`?

cljfmt/src/cljfmt/core.cljc
```diff
@@ -336,6 +363,9 @@
 (defmethod indenter-fn :block [sym context [_ idx]]
   (fn [zloc] (block-indent zloc sym idx context)))
 
+(defmethod indenter-fn :thread [_ _ _]
+  (constantly nil))
+
```
weavejester, line 368: I think we need to rethink how the indentation functions are applied, as the abstraction has clearly failed if we have an indenter function that does nothing.

lambdank:
No worries, I postponed this fix for several weeks.

I also thought about an approach, where the indentation happens by going through the AST recursively. 
As far as I can see, the indentation is only neighbors and parents, but not children. In a simplistic view, the node is also just impacted by nodes before, but that limits stuff like pair-alignment. This might be fixable by doing some sort of reverse-sweep with the context.
I just wanted to fix this issue with the least intrusive solution.

Feel free to reach out, if you get time at some point and envision a solution where I can help with.

PR #345, opened 2024-08-12
Don't rewrite `~ @foo` to `~@foo`

frenchy64:
The problem is `~ @foo` is being reformatted to `~@foo`, a different expression. `~,@foo` is used as the canonical form because it seems to have more emphasis and stability than `~ @foo` (just by my taste, lmk).

This decision has knock-on effects if there are are other forms between `~` and `@` which are demonstrated in the tests.

The implementation uses `transform` twice, but it should ideally just be one tree pass. Please let me know if there's a preferred way to do this.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -33,10 +33,6 @@
 (defn- transform [form zf & args]
   (z/root (apply zf (z/of-node form) args)))
 
-(defn- surrounding? [zloc p?]
-  (and (p? zloc) (or (nil? (z/left* zloc))
-                     (nil? (z/skip z/right* p? zloc)))))
-
 (defn root? [zloc]
   (nil? (z/up* zloc)))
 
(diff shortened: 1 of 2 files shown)


weavejester:
Thanks for the bug report, and the PR. Can you explain why you used a volatile? I don't understand why it's necessary. Also, why is `~ @foo` rewritten to `~,@foo`? Why not keep it as `~ @foo`?

frenchy64:
> Can you explain why you used a volatile? I don't understand why it's necessary.

Starting with the premise of using `transform` twice, I used a volatile as a performance optimization to only perform the second walk if absolutely necessary.

It would be better to have just one walk but I didn't know the best way to do this.

> Also, why is ~ @foo rewritten to ~,@foo? Why not keep it as ~ @foo?

This part is up to you.

My thought was: it's a really strange example of nested syntax that requires a space between the container and the child. I can't think of any other examples like it.

My idea with `~,@foo` was to make the disambiguation more explicit to draw attention to itself.

weavejester:
I think I'd like to know why cljfmt is removing the space in the first place. In the PR, code is changed for `insert-missing-whitespace`, presumably to re-insert the space that's removed by some other mechanism, but why not fix the code that causes it to remove the space in the first place?

frenchy64:
Ah, I misinterpreted `insert-missing-whitespace` as reinserting essential whitespace that was removed by cljfmt. I'll look into your suggestion.

frenchy64:
@weavejester seems to work!

weavejester:
Much better! I don't think I have any more notes. Could you squash your commits down?

frenchy64:
Done.

PR #346, opened 2024-08-20
Ignore metadata in front of first symbol when looking for matching rules

camsaul:
I could not get things like 

```clj
(^:once fn* [x y]
  (+ x y))
```

to format correctly despite having custom rules for `fn*`; after digging into things a bit I realized this gets parsed to a metadata node wrapping `fn*` and `token-value` doesn't work correctly in this case. Changed `token-value` to unwrap metadata nodes in this case which means the correct rule is now found and things indent according to the appropriate rule

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -243,7 +243,8 @@
        (some-> zloc z/down ns-token?)))
 
 (defn- token-value [zloc]
-  (when (token? zloc) (z/sexpr zloc)))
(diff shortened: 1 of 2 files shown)


weavejester:
Thanks for the commit! Can you change the commit message to:

```
Fix indentation for symbols preceded by metadata
```

This adds the keyword 'Fix' (useful for searching), and reduced the commit subject to 50 characters or less.

camsaul:
@weavejester done

weavejester:
Merged. Thanks for the work!

PR #347, opened 2024-08-20
Support specifying :default spec in custom indentation rules

camsaul:
Sometimes the default fuzzy rule for `^with-` of `[[:inner 0]]` is inappropriate, for example if you had a function like

```clj
(defn with-timezone [t zone]
  ...)
 
;;; this is a regular function and should follow `:default` list indentation rules
(with-timezone (java.time.OffsetDateTime/now)
               "US/Pacific")
```

This PR adds support for explicitly specifying `:default` formatting for a specific symbol

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -336,6 +336,11 @@
 (defmethod indenter-fn :block [sym context [_ idx]]
   (fn [zloc] (block-indent zloc sym idx context)))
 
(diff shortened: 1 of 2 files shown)


weavejester:
Thanks for the PR. This seems like a useful addition. Can you change the commit message to:

```
Add ':default' custom indentation rule
```

This puts it under 50 characters.

camsaul:
@weavejester done

weavejester:
Merged. Thanks for the work (again!)

PR #348, opened 2024-08-20
Fix indentation for symbols in reader conditionals

camsaul:
Things like

```clj
(#?(:clj potemkin.core/defprotocol+ :cljs defprotocol) MyProtocol
  ...)
```

were not indenting correctly despite having indentation specs defined for both `potemkin.core/defprotocol+` and `defprotocol`. 

This PR reworks `form-symbol` a little bit to return the first symbol it finds when it encounters `#?` forms -- in this case, `potemkin.core/defprotocol+` (meaning the form above will be formatted using that spec)

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -248,8 +248,30 @@
 (defn- reader-conditional? [zloc]
   (and (reader-macro? zloc) (#{"?" "?@"} (-> zloc z/down token-value str))))
 
+(defn- reader-conditional-form-symbol
+  "Given a zipper pointing to a reader conditional form like
+
+    #?(:clj potemkin.core/defprotocol+ :cljs defprotocol)
+
+  Return the first symbol inside it e.g.
+
+    potemkin.core/defprotocol+"
+  [zloc]
+  (when (reader-conditional? zloc)
+    ;; find the first keyword e.g. `:clj`
+    (when-let [key-loc (z/find (-> zloc z/down z/right z/down)
+                               z/right
+                               #(n/keyword-node? (z/node %)))]
+      ;; now move to the next non-whitespace/non-comment/non-metadata node after
+      ;; the keyword node e.g. `potemkin.core/defprotocol+`.
+      (when-let [value-loc (-> key-loc z/next skip-meta)]
+        (when (token? value-loc)
+          (z/sexpr value-loc))))))
+
 (defn- form-symbol [zloc]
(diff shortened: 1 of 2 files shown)


cljfmt/src/cljfmt/core.cljc
```diff
@@ -248,8 +248,30 @@
...
   (and (reader-macro? zloc) (#{"?" "?@"} (-> zloc z/down token-value str))))
 
+(defn- reader-conditional-form-symbol
+  "Given a zipper pointing to a reader conditional form like
+
+    #?(:clj potemkin.core/defprotocol+ :cljs defprotocol)
+
+  Return the first symbol inside it e.g.
+
+    potemkin.core/defprotocol+"
```
weavejester, line 258: Can you remove this docstring? Since this is a private function, we don't need to give it public documentation. Instead, perhaps improve the function name. Something like: `first-symbol-in-reader-conditional`.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -248,8 +248,30 @@
...
+  [zloc]
+  (when (reader-conditional? zloc)
+    ;; find the first keyword e.g. `:clj`
+    (when-let [key-loc (z/find (-> zloc z/down z/right z/down)
+                               z/right
+                               #(n/keyword-node? (z/node %)))]
```
weavejester, line 264: Maybe instead of the comment, we use a more descriptive function:

```clojure
(when-key [key-zloc (-> zloc z/down z/right z/down find-next-keyword)]
```

cljfmt/src/cljfmt/core.cljc
```diff
@@ -248,8 +248,30 @@
...
+      ;; now move to the next non-whitespace/non-comment/non-metadata node after
+      ;; the keyword node e.g. `potemkin.core/defprotocol+`.
```
weavejester, line 266: I don't think we need this comment, the code seems clear enough, and if it isn't, we should split it out into more functions. Ideally comments should explain "why" rather than "what".

weavejester (changes requested):
Thanks for the PR! I have some comments/changes after a review.

cljfmt/src/cljfmt/core.cljc
camsaul: done

cljfmt/src/cljfmt/core.cljc
camsaul: done

PR #350, opened 2024-08-21
Fix nested :inner specs

### LINKED ISSUE #349: `extend-protocol`/`defprotocol`/etc. incorrectly uses indentation specs for method names
There seems to be a problem with `[[:inner 1]] indentation handling when you have a indentation spec defined that matches the form symbol inside something like `extend-protocol` or `defprotocol`. E.g. if you have a config like

```clj
{:extra-indents '{my.namespace/with-x [[:block 0]]}}
```

Then a file like 

```clj
(ns my.namespace)

(defprotocol MyProtocol
  MyClass
  (with-x [this x]
    "with-x is a method"))

(extend-protocol MyProtocol
  MyClass
  (with-x [this x]
    (+ this x)))
```

Gets reformatted to

```clj
(ns my.namespace)

(defprotocol MyProtocol
  MyClass
  (with-x [this x]
          "with-x is a method"))

(extend-protocol MyProtocol
  MyClass
  (with-x [this x]
          (+ this x)))
```

It's using the `[[:block 0]]` spec for `my.namespace/with-x` rather than respecting the `[:inner 1]` rules specified for `defprotocol` and `extend-protocol`. 

Failing test:

```clj
(is (reformats-to?
     ["(ns my.namespace)"
      ""
      "(defprotocol MyProtocol"
      "MyClass"
      "(with-x [this x]"
      "\"with-x is a method\"))"
      ""
      "(extend-protocol MyProtocol"
      "MyClass"
      "(with-x [this x]"
      "(+ this x)))"]
     ["(ns my.namespace)"
      ""
      "(defprotocol MyProtocol"
      "  MyClass"
      "  (with-x [this x]"
      "    \"with-x is a method\"))"
      ""
      "(extend-protocol MyProtocol"
      "  MyClass"
      "  (with-x [this x]"
      "    (+ this x)))"]
     {:extra-indents '{my.namespace/with-x [[:block 0]]}}))
```

camsaul:
Fixes #349

After spending a few hours trying to figure out the best way to solve this problem without changing too much stuff I decided to changed the `indent` order code as follows:

- specs with multiple parts like `[[:block 1] [:inner 1]]` now get compiled to separate indenters for each part of the spec e.g. one for `[:block 1]` and one for `[:inner 1]`
- `:inner` specs with depth greater than zero are sorted preferentially with deeper depths sorted first. 

Previously in #349 the `sorted-indents` we see 

https://github.com/weavejester/cljfmt/blob/d84bd41b08e3e13662c328dd1345e930d6fd6001/cljfmt/src/cljfmt/core.cljc#L384

would have looked like 

```clj
[[my.namespace/with-x [[:block 0]]]
 [extend-protocol [[:block 1] [:inner 1]]]
```

This PR changes them so they now look like

```clj
[[extend-protocol [[:inner 1]]
 [my.namespace/with-x [[:block 0]]]
 [extend-protocol [[:block 1]]]
```

This ensures nested `:inner` indenters take precedence over general top-level indenters and `defprotocol`/`extend-protocol`/etc. are applied in a sane way.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -339,11 +339,25 @@
 (defn- make-indenter [[key opts] context]
   (apply some-fn (map (partial indenter-fn key context) opts)))
 
-(defn- indent-order [[key _]]
-  (cond
-    (and (symbol? key) (namespace key)) (str 0 key)
-    (symbol? key) (str 1 key)
-    (pattern? key) (str 2 key)))
+(defn- indent-order [[key [spec-type :as spec]]]
+  (let [depth*    (case spec-type
+                    :inner (let [[_inner depth] spec]
+                             (- 0 depth))
+                    0)
+        key-type* (cond
+                    (qualified-symbol? key) 0
+                    (simple-symbol? key)    1
+                    (pattern? key)          2)]
+    [depth* key-type* (str key)]))
+
+(defn- sorted-indents [indents]
+  (->> indents
+       (mapcat (fn [[k specs]]
+                 (for [spec specs]
+                   [k spec])))
+       (sort-by indent-order)
+       (map (fn [[k spec]]
+              [k [spec]]))))
 
 (defn- custom-indent [zloc indents context]
   (if (empty? indents)
@@ -381,7 +395,7 @@
    (indent form indents alias-map default-options))
   ([form indents alias-map opts]
    (let [ns-name (find-namespace (z/of-node form))
-         sorted-indents (sort-by indent-order indents)
+         sorted-indents (sorted-indents indents)
          context (merge (select-keys opts [:function-arguments-indentation])
                         {:alias-map alias-map
                          :ns-name ns-name})]
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -173,7 +173,40 @@
          ["(defrecord Foo [x]"
           "  Closeable"
           "  (close [_]"
-          "    (prn x)))"])))
+          "    (prn x)))"]))
+    (testing "nested rules like [:inner 1] (#349)"
+      (is (reformats-to?
+           ["(ns my.namespace)"
+            ""
+            "(defprotocol MyProtocol"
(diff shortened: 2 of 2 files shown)


weavejester:
What's the advantage of splitting up the indentation rules, compared to fixing the indentation ordering to prefer more 'outer' indentations?

camsaul:
Originally I thought keeping them as one rule it would potentially break stuff in weird edge cases , e.g. if you have 

```clj
letfn [[:block 0] [:inner 1]]
my.namespace/letfn [[:block 0]]
```

the stuff at depth one inside a `my.namespace/letfn` could be getting formatted as `[:inner 1]` but now that I squint at it a bit, it's a) a separate issue, b) splitting the rules up doesn't fix the problem, and c) a really contrived issue anyway.

I updated the PR so it keeps things as one rule and only updates the `indent-order` function

cljfmt/src/cljfmt/core.cljc
```diff
@@ -339,11 +339,21 @@
...
+                                     depth)
+                            0)))
+                   max
+                   0
+                   specs)
+        key-type* (cond
+                    (qualified-symbol? key) 0
+                    (simple-symbol? key)    1
+                    (pattern? key)          2)]
+    [(- max-depth) key-type* (str key)]))
```
weavejester, line 356: Perhaps instead:

```clojure
(defn- indent-order [[key specs]]
  (let [get-depth (fn [[type depth]] (if (= type :inner) depth 0))
        max-depth (transduce (map get-depth) max 0 specs)
        key-order  (cond
                    (qualified-symbol? key) 0
                    (simple-symbol? key)    1
                    (pattern? key)          2)]
    [(- max-depth) key-order (str key)]))

weavejester (changes requested):
Thanks for the PR. I only had one small suggestion around readability.

cljfmt/src/cljfmt/core.cljc
camsaul: done

weavejester:
Thanks for the update, and apologies for the delay in reviewing it. This past month has been rather busy for me.

PR #352, opened 2024-08-21
Fix formatting when requires have metadata

### LINKED ISSUE #351: Indentation rules break if namespace form in `:require` has metadata on it
When using `cljfmt` sorting ns `:require` forms with `:clj-kondo/ignore` comments doesn't really work. e.g.

```clj
(ns my.namespace
  (:require
   [a1.namespace]
   #_{:clj-kondo/ignore [:discouraged-namespace]}
   [a2.discouraged-namespace :as a2]))
```

gets sorted to 

```clj
(ns my.namespace
  (:require
   #_{:clj-kondo/ignore [:discouraged-namespace]}
   [a1.namespace]
   [a2.discouraged-namespace :as a2]))
```

to work around this issue I went ahead and changed the `:require` form from a comment to metadata, like this:

```clj
(ns my.namespace
  (:require
   [a1.namespace]
   ^{:clj-kondo/ignore [:discouraged-namespace]}
   [a2.discouraged-namespace] :as a2))
```

This keeps Kondo happy (Kondo ignore forms can be either comments or metadata) and `cljfmt` preserves the sort order... however namespace alias rules no longer work. E.g. if I have a rule like

```clj
a2.discouraged.namespace/my-do [[:inner 0]]
```

then something like 

```clj
(a2/my-do
  ...)
```

no longer formats correctly after changing the kondo ignore form from a `#_` comment to `^{}` metadata. 

I'm guessing there's an error in the logic that parses the `ns` require form to build the alias map that doesn't work correctly if there is metadata on one of the vectors in there -- probably since `:meta` is its own node type in rewrite-clj

camsaul:
Fixes #351

While working on this I tried to think of all the possible weird edge cases I could and found that it also broke in a few even weirder situations e.g. if you did something like put  metadata on or comments before the `x` in `[x :as y]` itself. I added a pretty comprehensive test makes sure things work as expected no matter what shenanigans you're up to in your `ns` form.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -566,12 +566,22 @@
           (= :as (z/sexpr zloc)))))
 
 #?(:clj
+   (defn- symbol-node? [zloc]
+     (some-> zloc z/node n/symbol-node?)))
+
+#?(:clj
+   (defn- leftmost-symbol [zloc]
+     (some-> zloc z/leftmost (z/find (comp symbol-node? skip-meta)))))
+
+#?(:clj
    (defn- as-zloc->alias-mapping [as-zloc]
      (let [alias             (some-> as-zloc z/right z/sexpr)
-           current-namespace (some-> as-zloc z/leftmost z/sexpr)
+           current-namespace (some-> as-zloc leftmost-symbol z/sexpr)
            grandparent-node  (some-> as-zloc z/up z/up)
            parent-namespace  (when-not (ns-require-form? grandparent-node)
-                               (first (z/child-sexprs grandparent-node)))]
+                               (when (or (z/vector? grandparent-node)
+                                         (z/list? grandparent-node))
+                                 (first (z/child-sexprs grandparent-node))))]
        (when (and (symbol? alias) (symbol? current-namespace))
          {(str alias) (if parent-namespace
                         (format "%s.%s" parent-namespace current-namespace)
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -257,6 +257,33 @@
          {:indents {'thing.core/defn [[:inner 0]]}
           #?@(:cljs [:alias-map {"t" "thing.core"}])})
         "applies custom indentation to namespaced defn")
+    (testing "handles metadata on or comments before forms inside ns :require list"
+      (doseq [ignore-str [""
+                          "^{:clj-kondo/ignore [:discouraged-namespace]} "
+                          "^:clj-kondo/ignore "
+                          "^{some-symbol another-symbol} "
+                          "#_{:clj-kondo/ignore [:discouraged-namespace]} "
+                          "#_:clj-kondo/ignore "
+                          "^tag "
+                          "#_old-thing "]
+              ns-vec-str [(str ignore-str "[thing.core :as t]")
+                          (str ignore-str "[thing [core :as t]]")
+                          (str ignore-str "(thing [core :as t])")
+                          (str "[" ignore-str "thing.core :as t]")
+                          (str ignore-str " [" ignore-str "thing.core :as t]")]
+              :let [ns-str (str "(ns example (:require " ns-vec-str "))")]]
+        (testing ns-str
+          (is (reformats-to?
+               [ns-str
+                ""
+                "(t/defn foo [x]"
+                "(+ x 1))"]
+               [ns-str
+                ""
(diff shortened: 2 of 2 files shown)


cljfmt/src/cljfmt/core.cljc
```diff
@@ -568,10 +568,43 @@
 #?(:clj
    (defn- as-zloc->alias-mapping [as-zloc]
      (let [alias             (some-> as-zloc z/right z/sexpr)
-           current-namespace (some-> as-zloc z/leftmost z/sexpr)
+           ;; `leftmost` might be an `uneval` node e.g. in
```
camsaul, line 571: I tried to write out the reasoning behind the changes here as opposed to in the PR, hope that's ok. Maybe it will save the next person who comes along some time so they don't have to work out what's going on the hard way

cljfmt/src/cljfmt/core.cljc
```diff
@@ -568,10 +568,16 @@
...
-           current-namespace (some-> as-zloc z/leftmost z/sexpr)
+           current-namespace (some-> as-zloc
+                                     z/leftmost
+                                     (z/find #(n/symbol-node?
+                                               (z/node (skip-meta %))))
+                                     z/sexpr)
            grandparent-node  (some-> as-zloc z/up z/up)
            parent-namespace  (when-not (ns-require-form? grandparent-node)
-                               (first (z/child-sexprs grandparent-node)))]
+                               (when (or (z/vector? grandparent-node)
```
camsaul, line 578: if we have something like

```clj
^{:clj-kondo/ignore [:discouraged-namespace} [my.namespace :as n]
```

then `grandparent-node` in this case is a `:meta` node so by adding the vector or string check we prevent that from accidentally returning an alias map like 

```clj
{"n" "{:clj-kondo/ignore [:discouraged-namespace]}.my.namespace"}
```

cljfmt/src/cljfmt/core.cljc
```diff
@@ -568,10 +568,16 @@
 #?(:clj
    (defn- as-zloc->alias-mapping [as-zloc]
      (let [alias             (some-> as-zloc z/right z/sexpr)
-           current-namespace (some-> as-zloc z/leftmost z/sexpr)
+           current-namespace (some-> as-zloc
+                                     z/leftmost
+                                     (z/find #(n/symbol-node?
+                                               (z/node (skip-meta %))))
+                                     z/sexpr)
```
camsaul, line 575: If you have something like 

```clj
[#_my.old.namespace my.namespace :as n]
```

then `z/leftmost` will be an `:uneval` node (see https://github.com/clj-commons/rewrite-clj/issues/70) so we need to skip it until we find the first **symbol** node, hence the use of `z/find`. I didn't use `token?` here since keywords are tokens too and we don't want it to accidentally return `:as` if it runs into some other issue.

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -257,6 +257,33 @@
          {:indents {'thing.core/defn [[:inner 0]]}
           #?@(:cljs [:alias-map {"t" "thing.core"}])})
         "applies custom indentation to namespaced defn")
+    (testing "handles metadata on or comments before forms inside ns :require list"
+      (doseq [ignore-str [""
```
camsaul, line 261: Hope using `doseq` to test different combinations of stuff here is ok. Writing out 40 separate test cases seemed a bit noisy

weavejester: That's fine, but why did you use `doseq` instead of `are`?

weavejester (changes requested):

cljfmt/src/cljfmt/core.cljc
```diff
@@ -565,13 +565,21 @@
...
+  (some-> zloc z/node n/symbol-node?))
+
+(defn- leftmost-symbol [zloc]
+  (some-> zloc z/leftmost (z/find (comp symbol-node? skip-meta))))
+
 #?(:clj
    (defn- as-zloc->alias-mapping [as-zloc]
      (let [alias             (some-> as-zloc z/right z/sexpr)
-           current-namespace (some-> as-zloc z/leftmost z/sexpr)
+           current-namespace (some-> as-zloc leftmost-symbol z/sexpr)
```
camsaul, line 577: Updated with suggestions

camsaul:
@weavejester, anything else you want me to change here?

weavejester:
No, that all looks good. Thanks for the PR.

Issue #222, opened 2021-04-08
can't format comment 


vinurs:
have the code like this 
```
(defn test
  []
;; this is a comment
  (+
   1 2))

```
but the comment line can not be formatted, it shows `All source files formatted correctly`

weavejester:
I'd be okay with a PR that makes formatting comments optional, though I don't think it should be the default. Comments aren't code, and shouldn't necessarily have the same indentation.

lread:
There is some good discussion on formatting comments in this existing issue: #112

PR #362, opened 2025-03-22
Optionally indent "code-fragment comments" according to the Clojure Style Guide

pbwolf:
Issue #112 asked about aligning ";;" comments like code and attracted some discussion. Issue #222 is a more focused request for indenting ";;" comments. The Clojure Style Guide says, "Write comments on a particular fragment of code before that fragment and aligned with it, using two semicolons".  In cljfmt's source code, "core.cljc" appears to follow this convention.

This PR indents code-fragment comments, optionally, when the new option `:indent-code-fragment-comments?` is true. (The name follows the Clojure Style Guide.) This PR also adds a corresponding command-line option, a mention in the README under "Formatting Options", and a test.

The option to indent code-fragment comments is *false by default*, as suggested by weavejester in a comment on #222.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,10 @@ In order to load the standard configuration file from Leiningen, add the
   the same as `:indents`, except that this will **append** to the
   default indents.
 
+* `:indent-line-comments?` -
+  true if cljfmt should align whole-line `;;` comments with the code.
+  Defaults to false.
+
 * `:alias-map` -
   a map of namespace alias strings to fully qualified namespace
   names. This option is unnecessary in most cases, because cljfmt
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -90,6 +90,9 @@
 (defn- comment? [zloc]
   (some-> zloc z/node n/comment?))
 
+(defn- line-comment? [zloc]
+  (and (comment? zloc) (re-matches #"(?s);;([^;].*)?" (z/string zloc))))
+
 (defn- comma? [zloc]
   (some-> zloc z/node n/comma?))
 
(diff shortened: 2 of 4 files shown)


cljfmt/src/cljfmt/core.cljc
```diff
@@ -145,14 +145,33 @@
...
-  (transform form edit-all should-unindent? z/remove*))
+;; Static comment: in the Clojure Style Guide,
+;; "heading comment" or "margin comment", vs. "code fragment comment"
+(defn- static-comment-next? [zloc]
+  (when-let [znext (-> zloc z/next* skip-whitespace)]
+    (when (comment? znext)
+      (let [text (str (z/node znext))
+            is-code-fragment-comment (and (str/starts-with? text ";;")
+                                          (not (str/starts-with? text ";;;")))]
+        (not is-code-fragment-comment)))))
```
weavejester, line 156: Rather than "static-comment", it might be better to call it "non-line-comment". So something like:

```clojure
(defn- line-comment? [zloc]
  (and (comment? zloc) (->> zloc z/node str (re-matches #";;[^;].*")))

(defn- non-line-comment-next?
  (not (-> zloc z/next* skip-whitespace line-comment?)))
```
```

File: README.md [hunk 2]
```diff
@@ -238,6 +238,10 @@ In order to load the standard configuration file from Leiningen, add the
   the same as `:indents`, except that this will **append** to the
   default indents.
 
+* `:indent-code-fragment-comments?` -
```
weavejester, line 241: In the Clojure Style Guide, these are also called "line comments", which is more concise, so let's use that.

weavejester (changes requested):
Thanks for the PR! I have a couple of suggestions, and could you also change the commit message so that it follows [the seven rules of a great Git commit message](https://chris.beams.io/posts/git-commit/#seven-rules), as per the contributing guidelines.

cljfmt/src/cljfmt/core.cljc
weavejester: Or instead of `non-line-comment-next?` it might be better to just have `line-comment-next?`, since there's a `not` in the later `if` expression anyway. No point in inverting then inverting again.

pbwolf: Thank you. I disentangled it slightly to emphasize the parallel with `(not (comment-next?...`

README.md
pbwolf: Done!

cljfmt/src/cljfmt/core.cljc
```diff
@@ -145,14 +145,32 @@
...
-(defn unindent [form]
-  (transform form edit-all should-unindent? z/remove*))
+;; Distinguishes line comments (;;), which may be indentable,
+;; from heading comments (;;;) and inline comments (;), which are not
+(defn- comment-next-other-than-line-comment? [zloc]
+  (when-let [znext (-> zloc z/next* skip-whitespace)]
+    (when (comment? znext)
+      (let [text (str (z/node znext))]
+        (or (not (str/starts-with? text ";;"))
+            (str/starts-with? text ";;;"))))))
```
weavejester, line 155: Could split out the `line-comment?` predicate? That way we have code that clearly and precisely defines what a line comment is. So something like:

```clojure
(defn- line-comment? [zloc]
  (and (comment? zloc) (re-matches #";;[^;].*" (z/string zloc))))

(defn- comment-next-other-than-line-comment? [zloc]
  (when-let [znext (-> zloc z/next* skip-whitespace)]
    (and (comment? znext) (not (line-comment? znext)))))
```

weavejester: Perfect, thanks!

pbwolf:
> For consistency, could you remove the backticks in the commit message? It's in plaintext format, rather than markdown.

Added a commit without the backticks.

weavejester:
Could you squash down your commits? You seem to have four commits with the same message.

pbwolf:
> Could you squash down your commits? You seem to have four commits with the same message.

Squashed

pbwolf:
Corrected

weavejester:
Merged! Thanks for the PR!

Issue #363, opened 2025-04-09
Can't get `:alias-map` to work


tlaundal:
Hi, while investigating betterthantomorrow/calva#2772, I found that I was not able to get `cljfmt` to use the `:alias-map` at all. I think this might be a PEBCAK, but I did try to follow [the documentation](https://github.com/weavejester/cljfmt/blob/master/docs/INDENTS.md#overview) without success.

As a small reproduction example, I've used this `.cljfmt.edn` file as config to format itself, but I also have the same problem in `.clj` files without a relevant import in the `ns` form.
```edn
{:extra-indents {repro.macro/example-macro [[:block 2]]}

 :alias-map {macro repro.macro}

 :test-code [;; This is formatted correctly
             (repro.macro/example-macro :first-arg second-arg
               (some long body))
             ;; The alias map is not applied here?
             (macro/example-macro :first-arg second-arg
                                  (some long body))]}
```

The above is how cljfmt formats the file. I used [standalone](https://github.com/weavejester/cljfmt?tab=readme-ov-file#standalone) cljfmt 0.13.0.

weavejester:
Thanks for the report. It looks like this is a bug.

The tests pass because they check for strings but not symbols, so it would work if you used:

```clojure
:alias-map {"macro" "repro.macro"}
```

However, that's not what the INDENTS.md docs say (thought the README is correct), and I can't think of a good reason not to allow symbols as well. My guess is that there's been a regression somewhere that wasn't caught by the tests.

tlaundal:
I've made #364 to update `INDENTS.md` with a correct example. I may come back to this later and try to fix the regression, but it's not top of my list.

PR #364, opened 2025-04-09
Use strings for `:alias-map` in indent docs

tlaundal:
Since only strings are supported for `:alias-map` (for now), the example should show strings.

As per https://github.com/weavejester/cljfmt/issues/363#issuecomment-2789383831


weavejester:
Fixed by e1047ecbcbb49f926bd31324338b67ccc6bb5e39

PR #367, opened 2025-08-21
Fix indentation when previous line contains multi-byte codepoint

### LINKED ISSUE #366: non-ASCII characters cause 𝕨𝕚𝕝𝕕 indentation
I was feeling a little 𝕨𝕚𝕝𝕕 so I wrote some code like this (elided version):

```clj
(defn- clause-with-options->legacy-MBQL
  [[tag options & args :as clause]]
  (loop [style (mbql.s/options-style tag), options options]
    (case style
      ::mbql.s/options-style.none (into [tag] args)
      ...
      ::mbql.s/options-style.𝕨𝕚𝕝𝕕 (if (> (count args) 2)
                                    (recur ::mbql.s/options-style.mbql-5 (options->legacy-MBQL options))
                                    (recur ::mbql.s/options-style.last-unless-empty options)))))
```

Emacs (`clojure-mode` and/or CIDER) seems to indent this code correctly but when I run it thru `cljfmt` it does something even 𝕨𝕚𝕝𝕕𝕖𝕣:

```clj
(defn- clause-with-options->legacy-MBQL
  [[tag options & args :as clause]]
  (loop [style (mbql.s/options-style tag), options options]
    (case style
      ::mbql.s/options-style.none (into [tag] args)
      ...
      ::mbql.s/options-style.𝕨𝕚𝕝𝕕 (if (> (count args) 2)
                                        (recur ::mbql.s/options-style.mbql-5 (options->legacy-MBQL options))
                                        (recur ::mbql.s/options-style.last-unless-empty options)))))
```

As you can see it's incorrectly indented 4 spaces too much. I'm guessing this is almost certainly because while `𝕨𝕚𝕝𝕕` has 4 characters (4 Unicode code points at least), it has 8 bytes:

```clj
(count "𝕨𝕚𝕝𝕕")
;; => 8

(let [s "𝕨𝕚𝕝𝕕"]
  (.codePointCount s 0 (count s)))
;; => 4
```

Now this is probably a dumb example but I think it does expose a genuine bug in cljfmt where multi-byte codepoints cause incorrect formatting. I will need to dig into the code but hopefully it's a simple fix to update things to look at `.codePointCount` instead of the (byte) `count`/string length.

camsaul:
Fixes #366

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -191,11 +191,12 @@
         (recur p (cons (start-element (n/tag (z/node p))) worklist))
         (apply str worklist)))))
 
-(defn- last-line-in-string [^String s]
+(defn- last-line-in-string ^String [^String s]
   (subs s (inc (.lastIndexOf s "\n"))))
 
 (defn- margin [zloc]
-  (-> zloc prior-line-string last-line-in-string count))
+  (let [last-line (-> zloc prior-line-string last-line-in-string)]
+    (.codePointCount last-line 0 (.length last-line))))
 
 (defn- whitespace [width]
   (n/whitespace-node (apply str (repeat width " "))))
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -2009,3 +2009,20 @@
            ["(when 42  ;;;"
             "  :a)"]
            {:indent-line-comments? true})))))
+
+(deftest test-multibyte-codepoints
+  (is (reformats-to?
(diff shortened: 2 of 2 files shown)


weavejester:
Thanks for the fix! Can you change the commit message to:

```
Fix indent when line contains multi-byte codepoint
```

This ensures that the commit title is 50 characters or fewer. Once that's done, I'll merge it in.

weavejester:
Also, I think using `(.length last-line)` might be a little faster than `(count last-line)`, now that last-line has a String type hint.

camsaul:
@weavejester done

PR #375, opened 2025-10-20
Remove blank lines in forms

### LINKED ISSUE #374: Support removing blank lines in definition forms
I really really want something that will automatically fix code that doesn't follow https://guide.clojure.style/#no-blank-lines-within-def-forms. We'd need a way to exempt certain pairwise forms like `cond` or `let`, we'd want to make this extensible for non-core things like [`better-cond`](https://github.com/Engelberg/better-cond) that we also wanted to exempt. I think that would need an extra key in `.cljfmt.edn`.

Does this all make sense and sound like something you'd accept? I'm willing to submit a PR for this

camsaul:
Resolves #374

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,9 +247,22 @@ In order to load the standard configuration file from Leiningen, add the
   true if cljfmt should align the keys and values of maps such that they
   line up in columns. Defaults to false. **Experimental.**
 
+* `:blank-line-forms` - map of symbols that tell cljfmt which forms are allowed
+  to have blank lines inside of them. The value may be either `:all`, which
+  means blank lines are allowed between all elements in the form, e.g. `{'cond
+  :all}` to allow blank lines between any of the elements inside `cond`; or it
+  may be a set of element indexes that are allowed to have blank lines, e.g.
+  `{'let #{0}}`, to allow blank lines in the binding of a `let` form. This
+  option will **replace** the default blank line forms; use `:extra-blank-line`
+  forms to add additional ones. Used by `:remove-blank-lines-in-forms?`.
+  **Experimental.**
+
 * `:extra-aligned-forms` -
   the same as `:aligned-forms`, except that this will **append** to the
   default aligned forms. **Experimental.**
+
+* `:extra-blank-line-forms` - the same as `:blank-line-forms`, except that this
+  will **append** to the default blank line forms. **Experimental.**
 
 * `:extra-indents` -
   the same as `:indents`, except that this will **append** to the
@@ -280,6 +293,12 @@ In order to load the standard configuration file from Leiningen, add the
   a map of var symbols to indentation rules, i.e. `{symbol [& rules]}`.
   See [INDENTS.md][] for a complete explanation. This will **replace**
   the default indents.
+
+* `:remove-blank-lines-in-forms?` - whether to remove blank lines inside forms
+  per the [community style recommendation][no-blank-lines]. By default, this
+  does not remove blank lines in pairwise constructs like `cond` or within
+  binding vectors; configure these with `:blank-line-forms` and/or
+  `:extra-blank-line-forms`. Defaults to false. **Experimental**.
 
 * `:remove-consecutive-blank-lines?` -
   true if cljfmt should collapse consecutive blank lines. This will
@@ -315,6 +334,7 @@ In order to load the standard configuration file from Leiningen, add the
 
 [indents.md]: docs/INDENTS.md
 [community style recommendation]: https://guide.clojure.style/#one-space-indent
+[no-blank-lines]: https://guide.clojure.style/#no-blank-lines-within-def-forms
 
 ### Runtime Options
 
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -358,18 +358,24 @@
 (def default-aligned-forms
   (read-resource "cljfmt/aligned_forms/clojure.clj"))
 
+(def blank-line-forms
+  (read-resource "cljfmt/blank_line_forms/clojure.clj"))
+
 (def default-options
   {:alias-map                             {}
    :align-binding-columns?                false
    :align-map-columns?                    false
    :aligned-forms                         default-aligned-forms
+   :blank-line-forms                      blank-line-forms
    :extra-aligned-forms                   {}
+   :extra-blank-line-forms                {}
    :extra-indents                         {}
    :function-arguments-indentation        :community
    :indent-line-comments?                 false
    :indentation?                          true
    :indents                               default-indents
    :insert-missing-whitespace?            true
+   :remove-blank-lines-in-forms?          false
    :remove-consecutive-blank-lines?       true
    :remove-multiple-non-indenting-spaces? false
    :remove-surrounding-whitespace?        true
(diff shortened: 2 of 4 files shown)


camsaul:
> Is there any reason to limit this to definition forms?

I guess it's not really limited to definition forms, it removes blank lines anywhere except the top level, I just wasn't sure what to call this so I tried to match the naming of the rule in the Community Style Guide https://guide.clojure.style/#no-blank-lines-within-def-forms . But that's not quite the same as what this does so I'll change it

> what if we narrowed down where the blank lines were allowed

For binding vectors this builds on the recently-added `:[extra-]binding-forms` stuff, so for `let` currently just the binding vector can keep extra newlines, and for stuff like `cond` it's limited to the top-level of the form instead (nested forms inside `cond` still get considered separately). I think there's some advantage to this because you don't need to specify the binding forms for everything twice if you're also using `:align-binding-columns?`. Does that work? I can tweak it if you still want something more fine-grained

camsaul:
Updated the name of this to make it clearer that it works on all forms and not just `def` and friends

camsaul:
Makes sense. I'll rework this PR later today

cljfmt/src/cljfmt/core.cljc
```diff
@@ -681,6 +687,43 @@
...
+         (z/list? (z/up (z/up zloc)))
+         (let [index           (dec (index-of (z/up zloc)))
+               rule            (blank-line-rule
+                                (z/up zloc)
+                                blank-line-forms
+                                context)
+               allowed-indexes (if (set? rule)
+                                 rule
+                                 #{})]
+           (not (allowed-indexes index))))))
```
weavejester: This doesn't seem like it'll work for an arbitrary nesting depth. What we want to do is determine whether the zloc represents multiple newlines, and then iterate up the tree until we hit root (`nil`), checking if any containing form is also a matching form.

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -1157,6 +1157,119 @@
...
+                    "  (:require [better-cond.core :as b]))"
+                    ""
+                    "(b/cond*"
+                    "  (map? x)"
+                    "  :map"
+                    ""
+                    "  (sequential? x)"
+                    "  :seq)"]]
+          (is (reformats-to? form form {:remove-blank-lines-in-forms? true
+                                        :extra-indents                '{better-cond.core/cond* [[:block 0]]}
```
weavejester: Keep lines to 80 characters or less, even in tests.

README.md
```diff
@@ -281,6 +294,13 @@ In order to load the standard configuration file from Leiningen, add the
   See [INDENTS.md][] for a complete explanation. This will **replace**
   the default indents.
 
+* `:remove-blank-lines-in-forms?` - whether to remove blank lines inside forms
+  per the [community style
+  recommendation](https://guide.clojure.style/#no-blank-lines-within-def-forms).
```
weavejester: Use an external URL to this link, like `[community style recommendation][no-blank-lines]`.

weavejester (changes requested):
Thanks for the PR. I've reviewed the code and have some suggestions/improvements.

camsaul:
👍  I'll probably get around to fixing this up on Monday and then let you know when it's ready

cljfmt/src/cljfmt/core.cljc
```diff
@@ -702,6 +709,24 @@
       (edit-all non-indenting-whitespace? unalign-from-space z/right*)
       z/root))
 
+(defn- blank-line-in-form? [zloc blank-line-forms context]
+  (and (z/linebreak? zloc)
+       (> (count-newlines zloc) 1)
+       (some (fn [zloc*]
+               (and (z/list? (z/up zloc*))
+                    (not (matching-form? zloc* blank-line-forms context))))
+             [zloc (z/up zloc)])))
```
weavejester: Could you explain why we need to check `(z/up zloc)` and `(z/up (z/up zloc))`?

camsaul: `(z/up zloc)`  ends up matching `:all` rules in cases like this 

```clj
(cond
  a
  b
  <this-node-right-here>
  c
  d)
```

while `(z/up (z/up zloc))`  ends up matching the set-of-indexes rules (e.g. `#{0}`) in cases like this

```clj
(let [a
      1
      <this-node-right-here>
      b
      2]
  ...)
```

In the first case the form that the rules match (`cond`) is the parent of the newline node while in the second case the form (`let`) is the grandparent of the newline node.

I can move the logic for checking at two different levels into `matching-form?` itself if you'd like. I was trying to make things work using your suggested changes exactly but I think this change would make things cleaner

cljfmt/src/cljfmt/core.cljc
```diff
@@ -674,17 +680,24 @@
...
+                 (fn [[k indexes]]
+                   (and (form-matches-key? (z/down zloc) k context)
+                        (= :all indexes)))
+
+                 (z/list? (z/up zloc))
+                 (fn [[k indexes]]
+                   (and (set? indexes)
+                        (form-matches-key? zloc k context)
+                        (contains? (set indexes) (dec (index-of zloc))))))]
+    (some f aligned-forms)))
```
weavejester: I think we can make this a touch clearer and more efficient by splitting this into two functions:

```clojure
(defn- matching-form-index? [zloc [k indexes] context]
  (if (= :all indexes)
    (and (z/list? zloc)
         (form-matches-key? (z/down zloc) k context))
    (and (z/list? (z/up zloc))
         (form-matches-key? zloc k context)
         (contains? (set indexes) (dec (index-of zloc))))))

(defn- matching-form? [zloc form-indexes context]
  (and (or (z/list? zloc) (z/list (z/up zloc)))
       (some #(matching-form-index? zloc % context) form-indexes)))
```

So we have a quick check first, just to see if we're in a list or a direct child of a list, then we iterate through the form indexes to find a match.

weavejester (changes requested):
Thanks for your work on this. I have a couple of changes which I've added in comments, but otherwise I think we're pretty much there.

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -1188,6 +1188,123 @@
...
+    (let [form ["(def x 1)"
+                ""
+                "(def x 2)"]]
+      (is (reformats-to? form form {:remove-blank-lines-in-forms? true}))))
+  (testing "Ignore blank lines when parent and/or grandparent forms are not lists"
+    (let [form ["[:a"
+                " [:b"
+                ""
+                "  :c]]"]]
+      (is (reformats-to? form form {:remove-blank-lines-in-forms? true})))))
```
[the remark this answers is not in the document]
camsaul: I wanted to error on the side of formatting less stuff rather than more since things like vector or map literals aren't currently configurable. This is maybe not a great example but if I changed it to something like the map example in my previous comment then it would make a little more sense

cljfmt/src/cljfmt/core.cljc
```diff
@@ -702,6 +715,23 @@
       (edit-all non-indenting-whitespace? unalign-from-space z/right*)
       z/root))
 
+(defn- blank-line-in-form? [zloc blank-line-forms context]
+  (and (z/linebreak? zloc)
+       (> (count-newlines zloc) 1)
+       (not (root? (z/up zloc)))
+       (some z/list? [(z/up zloc) (z/up (z/up zloc))])
+       (not (matching-form? (z/up zloc) blank-line-forms context))))
```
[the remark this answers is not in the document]
camsaul: yeah I think this makes sense for now

weavejester:
I've released 0.15.6 with this feature. Thanks again for your work on this.

PR #381, opened 2025-12-10
Fix #378: Add :blank-lines-separate-alignment? option to separate alignment groups 

### LINKED ISSUE #378: Align vertical forms grouped by non-whitespace lines
The new features `:align-binding-columns?` `:align-form-columns?` and `:align-map-columns?` are a nice start, however unlike [CIDER's `clojure-align`](https://github.com/clojure-emacs/clojure-mode/blob/master/CHANGELOG.md?plain=1#L327) function, let and map forms with whitespace separating logical groups do not use variable indentation.

Actual
--
```clojure
(let [short-var                                        true
      much-longer-var-name-that-should-force-alignment true

      some-map                                         {:a            1
                                                        :alphabetical 2}]
    ...)
```

Expected
--
```clojure
(let [short-var                                        true
      much-longer-var-name-that-should-force-alignment true

      some-map {:a            1
                :alphabetical 2}]
    ...)
```

jramosg:
feat: add option to treat blank lines as alignment group separators

Introduce the :blank-lines-separate-alignment? option to control
alignment behavior in maps and aligned forms. When enabled, cljfmt
treats blank lines as boundaries for alignment groups, aligning
columns independently within each group separated by blank lines.
The CLI now supports --[no-]blank-lines-separate-alignment, and
the README documents the feature as experimental.

This allows for more granular control over column alignment,
especially in forms where logical groupings are separated by
blank lines.

Closes #378

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -374,6 +374,7 @@
    :align-single-column-lines?            false
    :aligned-forms                         default-aligned-forms
    :blank-line-forms                      blank-line-forms
+   :blank-lines-separate-alignment?       false
    :extra-aligned-forms                   {}
    :extra-blank-line-forms                {}
    :extra-indents                         {}
@@ -627,16 +628,6 @@
   (and (line-break? (skip-whitespace-and-commas (z/right* zloc) z/right*))
        (line-break? (skip-whitespace-and-commas (z/left* zloc) z/left*))))
 
-(defn- max-column-end-position [zloc col align-single-column-lines?]
-  (reduce-columns zloc
-                  (fn [zloc c max-pos]
-                    (if (and (= c col)
-                             (or align-single-column-lines?
-                                 (not (single-column-line? zloc))))
-                      (max max-pos (node-end-position zloc))
-                      max-pos))
-                  0))
-
 (defn- node-str-length [zloc]
   (-> zloc z/node n/string count))
 
@@ -681,24 +672,54 @@
           zloc))
       zloc)))
 
-(defn- align-one-column [zloc col align-single-column-lines?]
+(defn- end-of-column-group? [zloc]
+  (line-break? zloc) (> (count-newlines zloc) 1))
+
+(defn- find-start-of-column-group [zloc]
+  (z/skip z/left* #(some-> % z/left* end-of-column-group? not) zloc))
+
+(defn- reduce-column-group [zloc f init]
+  (loop [zloc (find-start-of-column-group zloc)
+         col 0, acc init]
+    (let [zloc (skip-whitespace-and-commas zloc z/right*)]
+      (if (or (nil? zloc) (end-of-column-group? zloc))
+        acc
+        (if (line-break? zloc)
+          (recur (z/right* zloc) 0 acc)
+          (recur (z/right* zloc) (inc col) (f zloc col acc)))))))
+
+(defn- column-start-position [zloc col opts]
+  (let [reduce-fn (if (:blank-lines-separate-alignment? opts)
+                    reduce-column-group
+                    reduce-columns)
+        maximizer (fn [zloc c max-pos]
+                    (if (and (= c (dec col))
+                             (or (:align-single-column-lines? opts)
+                                 (not (single-column-line? zloc))))
+                      (max max-pos (node-end-position zloc))
+                      max-pos))]
+    (inc (reduce-fn zloc maximizer 0))))
+
+(defn- align-one-column
+  [zloc col {:keys [blank-lines-separate-alignment?] :as opts}]
   (if-some [zloc (z/down zloc)]
-    (let [start-position (inc (max-column-end-position
-                               zloc (dec col) align-single-column-lines?))]
-      (z/up (edit-column zloc col #(pad-to-position % start-position))))
+    (let [start-position-fn (if blank-lines-separate-alignment?
+                              #(column-start-position % col opts)
+                              (constantly
+                               (column-start-position zloc col opts)))]
+      (z/up (edit-column zloc col #(pad-to-position % (start-position-fn %)))))
     zloc))
 
-(defn- align-columns [zloc align-single-column-lines?]
-  (reduce #(align-one-column %1 %2 align-single-column-lines?)
+(defn- align-columns [zloc opts]
+  (reduce #(align-one-column %1 %2 opts)
           zloc
           (-> zloc z/down count-columns range rest)))
 
 (defn align-map-columns
   ([form]
    (align-map-columns form default-options))
-  ([form {:keys [align-single-column-lines?]}]
-   (let [align #(align-columns % align-single-column-lines?)]
-     (transform form edit-all z/map? align))))
+  ([form opts]
+   (transform form edit-all z/map? #(align-columns % opts))))
 
 (defn- matching-form-index? [zloc [k indexes] context]
   (if (= :all indexes)
@@ -715,19 +736,18 @@
 (defn align-form-columns
   ([form aligned-forms alias-map]
    (align-form-columns form aligned-forms alias-map default-options))
-  ([form aligned-forms alias-map {:keys [align-single-column-lines?]}]
+  ([form aligned-forms alias-map opts]
    (let [ns-name  (find-namespace (z/of-node form))
          context  {:alias-map alias-map, :ns-name ns-name}
-         aligned? #(matching-form? % aligned-forms context)
-         align    #(align-columns % align-single-column-lines?)]
-     (transform form edit-all aligned? align))))
+         aligned? #(matching-form? % aligned-forms context)]
+     (transform form edit-all aligned? #(align-columns % opts)))))
 
 (defn realign-form
   "Realign a rewrite-clj form such that the columns line up into columns."
   ([form]
    (realign-form form default-options))
-  ([form {:keys [align-single-column-lines?]}]
-   (-> form z/of-node (align-columns align-single-column-lines?) z/root)))
+  ([form opts]
+   (-> form z/of-node (align-columns opts) z/root)))
 
 (defn- unalign-from-space [zloc]
   (pad-node (z/right* zloc) (- 1 (node-str-length zloc))))
--- a/README.md
+++ b/README.md
@@ -263,6 +263,12 @@ In order to load the standard configuration file from Leiningen, add the
   padding in other lines. Only applies when `:align-form-columns?` or  
   `:align-map-columns?` is true. Defaults to false. **Experimental.**  
 
+* `blank-lines-separate-alignment` -
+  true if cljfmt should treat blank lines as separators when
+  aligning columns. When enabled, alignment groups are separated by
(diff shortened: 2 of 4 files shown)


kevinmershon:
@jramosg I don't know you but I love you

jramosg:
Thanks for the feedback! I will try to fix both and I will add them or similar examples to tests. 😊

cljfmt/src/cljfmt/core.cljc
```diff
@@ -675,17 +676,75 @@
...
+               (align-one-column-with-groups zloc col))
+             zloc
+             (column-indexes zloc))
+     (reduce (fn [zloc col]
+               (align-one-column zloc col))
+             zloc
+             (column-indexes zloc)))))
+
+(defn align-map-columns [form blank-lines-separate-alignment?]
+  (transform form edit-all z/map? #(align-columns % blank-lines-separate-alignment?)))
```
weavejester: This could be a little more concisely expressed as:

```clojure
(defn- align-columns [zloc alignf]
  (reduce alignf zloc (-> zloc z/down count-columns range rest)))

(defn align-map-columns [form blank-lines-separate-alignment?]
  (let [alignf (if blank-lines-separate-alignment?
                 align-one-column
                 align-one-column-with-groups))]
    (transform form edit-all z/map? #(align-columns % alignf))))
```

kevinmershon:
@jramosg it works perfectly and you are a saint

weavejester:
> However, I couldn't apply the suggested simplification using edit-column with a start-pos function because:
> 
> Multiple groups require multiple calculations: Each group separated by blank lines needs its own independent alignment position edit-column doesn't respect group boundaries: It processes the entire structure sequentially without stopping at blank lines Per-element calculation would be incorrect: Calling max-column-group-end-position for each element would recalculate from that element's position forward, giving wrong results for subsequent elements in the same group

Sorry, I should have been clearer. I'm thinking of a function like this:

1. Skips left until closest blank line, comment or the start of the form (i.e finds the group start).
2. Then reduces right over the form until the next blank line, comment or end of form (i.e. until the group end).

So roughly:

```clojure
(defn- reduce-column-group [zloc f init]
  (loop [zloc (find-start-of-column-group zloc)
         col 0, acc init]
    (let [zloc (skip-to-linebreak-or-element zloc)]
      (if (end-of-column-group? zloc)
        acc
        (if (line-break? zloc)
          (recur (z/right* zloc) 0 acc))
          (recur (z/right* zloc) (inc col) (f zloc col acc)))))))
```

Once we have this, we can then write a function to calculate the max width for the current group that the zloc is in, regardless of where in that group the zloc is.

```clojure
(defn- column-end-maximizer [col]
  (fn [zloc c max-pos]
    (if (= c col)
      (max max-pos (node-end-position zloc))
      max-pos)) 

(defn- max-column-end-position [zloc col]
  (reduce-columns zloc (column-end-maximizer col) 0))

(defn max-column-group-end-position [zloc col]
  (reduce-column-group zloc (column-end-maximizer col) 0))
```

weavejester:
Thanks! It's important to ensure PRs are focused so they're easy to understand when I come back to them in a decade :smiley:

jramosg:
Hello @weavejester
I've updated this PR to ensure full compatibility and proper integration with the `align-single-column-lines?` feature. The changes refine the alignment logic to handle combinations of both options correctly.

The PR is now ready for review. Let me know if you'd like any adjustments or have questions about the implementation.

Thank you

jramosg:
Hello @weavejester 

No rush at all, I understand you may be busy. I just wanted to check whether it’s something you might be planning to review, or if there’s anything I could do to help move it forward.

Thanks for your work on the project.

weavejester:
Yes, apologies for the delay - January's been a busy month for me so I haven't had a lot of time for open source work.

I've re-opened the PR as I assume closing it was a mistake?

**Edit:** or perhaps not a mistake? All the commits seem to have gone?

weavejester:
I've taken an initial look over the rebased code and I think my [previous comment](https://github.com/weavejester/cljfmt/pull/381#issuecomment-3639464796) still applies. We could cut down this code a lot if we reused existing functions, rather than create new loops over the zipper.

Fundamentally, the change we're making is determining where the new column starts. Do we take the maximum length of the previous column in the entire form, or do we take the maximum length of the current group?

So, if we take a look at the current `align-one-column` function:

```clojure
(defn- align-one-column [zloc col align-single-column-lines?]
  (if-some [zloc (z/down zloc)]
    (let [start-position (inc (max-column-end-position
                               zloc (dec col) align-single-column-lines?))]
      (z/up (edit-column zloc col #(pad-to-position % start-position))))
    zloc))
```

Currently we calculate a constant `start-position` for the entire form, but we could generalize that by making it a function. For example:

```clojure
(defn- align-one-column
  [zloc col {:keys [blank-lines-separate-alignment?] :as opts}]
  (if-some [zloc (z/down zloc)]
    (let [start-position (if blank-lines-separate-alignment?
                           (constantly (column-start-position zloc col opts))
                           #(column-group-start-position % col opts))]
      (z/up (edit-column zloc col #(pad-to-position % (start-position %)))))
    zloc))
```

When we're not using groups, we calculate the start position for the whole form and create a function that constantly returns this number; when we are using groups, we work out the start position for the group at the current zloc.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -680,24 +679,79 @@
...
+  (loop [zloc (find-start-of-column-group zloc)
+         col 0
+         acc init]
+    (if-some [zloc (skip-whitespace-and-commas zloc z/right*)]
+      (if (end-of-column-group? zloc)
+        acc
+        (if (line-break? zloc)
+          (recur (z/right* zloc) 0 acc)
+          (recur (z/right* zloc) (inc col) (f zloc col acc))))
+      acc)))
```
weavejester: This could be a little more concise. There's no need to use `if-some` if `end-of-column-group?` accounts for `nil`.

```clojure
(defn- reduce-column-group [zloc f init]
  (loop [zloc (find-start-of-column-group zloc)
         col 0, acc init]
    (let [zloc (skip-whitespace-and-commas zloc z/right*)]
      (if (end-of-column-group? zloc)
        acc
        (if (line-break? zloc)
          (recur (z/right* zloc) 0 acc)
          (recur (z/right* zloc) (inc col) (f zloc col acc)))))))
```

cljfmt/src/cljfmt/core.cljc
```diff
@@ -680,24 +679,79 @@
...
+         maximizer (fn [zloc c max-pos]
+                     (if (and (= c col)
+                              (not (and (pos? c)
+                                        (preceded-by-linebreak? zloc)))
+                              (not (comment? zloc))
+                              (or align-single-column-lines?
+                                  (not (single-column-line? zloc))))
+                       (max max-pos (node-end-position zloc))
+                       max-pos))]
+     (reduce-column-group zloc maximizer 0))))
```
weavejester: Could you explain why the maximizer functions are different?

cljfmt/src/cljfmt/core.cljc
```diff
@@ -680,24 +679,79 @@
...
+(defn- end-of-column-group? [zloc]
+  (or (nil? zloc)
+      (has-blank-line-after? zloc)))
+
+(defn- find-start-of-column-group [zloc]
+  (loop [z zloc]
+    (let [prev (when z (z/left* z))]
+      (if (or (nil? prev) (has-blank-line-after? prev))
+        z
+        (recur prev)))))
```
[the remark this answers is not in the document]
jramosg: In `end-of-colum-group?`  `(pos? (count-newlines zloc))` could be a normal line break within the same group, so we still need `(> (count-newlines zloc) 1)`.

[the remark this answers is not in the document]
jramosg: also it has not be nil, so
```clojure
(defn- find-start-of-column-group [zloc]
  (z/skip z/left*
          #(when-let [prev (z/left* %)]
             (not (end-of-column-group? prev)))
          zloc))
```

cljfmt/src/cljfmt/core.cljc
jramosg: we must check for nil, otherwise we are in an infitie loop.  where does check `end-of-column-group?` for nil?
```clojure
(defn- end-of-column-group? [zloc]
  (line-break? zloc) (> (count-newlines zloc) 1))
```

cljfmt/src/cljfmt/core.cljc
jramosg: i don't actually needed, changed

cljfmt/src/cljfmt/core.cljc
weavejester: Right, so we'd want something like:

```clojure
(defn- end-of-column-group? [zloc]
  (or (nil? zloc) (and (line-break? zloc) (> (count-newlines zloc) 1))))
```

Or perhaps it would be better to place the `nil` check in `reduce-column-group`:

```clojure
(defn- reduce-column-group [zloc f init]
  (loop [zloc (find-start-of-column-group zloc)
         col 0, acc init]
    (let [zloc (skip-whitespace-and-commas zloc z/right*)]
      (if (or (nil? zloc) (end-of-column-group? zloc))
        acc
        (if (line-break? zloc)
          (recur (z/right* zloc) 0 acc)
          (recur (z/right* zloc) (inc col) (f zloc col acc)))))))
```

That might be better, as `nil` isn't technically the end of a column group, at least not in isolation.

jramosg: :ok_hand: changed

cljfmt/src/cljfmt/core.cljc
```diff
@@ -681,24 +680,58 @@
...
+
+(defn- column-start-position [zloc col opts]
+  (let [reduce-fn (if (:blank-lines-separate-alignment? opts)
+                    reduce-column-group
+                    reduce-columns)
+        maximizer (fn [zloc c max-pos]
+                    (if (and (= c col)
+                             (not (and (pos? c)
+                                       (preceded-by-linebreak? zloc)))
+                             (not (comment? zloc))
```
weavejester: Could you explain these changes in more detail? I don't understand the purpose of these additions.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -681,24 +672,55 @@
...
+                      max-pos))]
+    (inc (reduce-fn zloc maximizer 0))))
+
+(defn- align-one-column
+  [zloc col {:keys [blank-lines-separate-alignment?] :as opts}]
   (if-some [zloc (z/down zloc)]
-    (let [start-position (inc (max-column-end-position
-                               zloc (dec col) align-single-column-lines?))]
-      (z/up (edit-column zloc col #(pad-to-position % start-position))))
+    (let [start-position-fn (let [col (dec col)]
```
weavejester: This `(dec col)`, shouldn't it be in `column-start-position`? It looks like currently `column-start-position` actually finds the *end* position of the column (plus one).

PR #382, opened 2025-12-10
Fix #380: Preserve line breaks after #_ uneval tags to prevent moving…

### LINKED ISSUE #380: Keep the `#_` uneval tag untouched
Hi! This is crosspost from a Slack channel:

How can I control the `#_`  tag? Imagine a have a h-u-u-ge form that I'd like to ignore. For this, I just put `#_` on a previous line:

```
#_
(defn foobar []
 (do-this)
 (let [...]
    (with-blah []
      ...)))
```

But `cljfmt` corrects it in this way:

```
#_(defn foobar []
   (do-this)
   (let [...]
      (with-blah []
        ...)))
```

This is quite annoying because it leads to vast git changes. I tried to find an option to disable it but could not. Is there any?

jramosg:
Fix #380: Preserve line breaks after #_ uneval tags to prevent moving them to same line as following form

- Add uneval? helper function to identify uneval nodes
- Modify missing-whitespace? to protect linebreaks after uneval
- Update surrounding-whitespace? to prevent removal of linebreaks after uneval when it's the first child
- Handle uneval children indentation in indent-amount (no indent when on own line, margin+1 when inline)
- Adjust list-indent to account for uneval in argument positioning
- Add comprehensive tests covering inline vs own-line uneval, different indentation styles, multiple unevals, and various data structures"

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -238,22 +238,20 @@
     :zprint (zprint-two-space-list-indent? zloc)))
 
 (defn- first-form-in-line? [zloc]
-  (and (some? zloc)
-       (if-let [zloc (z/left* zloc)]
-         (if (space? zloc)
-           (recur zloc)
-           (or (z/linebreak? zloc) (comment? zloc)))
-         true)))
+  (if-let [zloc (z/left* zloc)]
+    (if (space? zloc)
+      (recur zloc)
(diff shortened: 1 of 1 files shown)


weavejester:
Thanks for the PR. However, I think this change could do with a little more discussion so we can confirm what the purpose of this change is, and what should be done to fix it. I'll add some comments to the original issue.

PR #383, opened 2025-12-11
Add align-single-column-lines? option

jramosg:
Add align-single-column-lines? option
Introduce a new configuration option `:align-single-column-lines?`
to control column alignment behavior in maps and forms. When set
to false (the default), lines with wrapped values are excluded
from alignment calculations, preventing excessive horizontal
padding in other lines. When true, all lines are considered for
alignment.

This addresses the issue where forms with multi-line values
would cause unwanted spacing in aligned columns, improving
readability and formatting consistency.

Resolves: https://github.com/weavejester/cljfmt/issues/384

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -107,8 +107,9 @@
 (defn- skip-whitespace [zloc]
   (z/skip z/next* space? zloc))
 
-(defn- skip-whitespace-and-commas [zloc]
-  (z/skip z/next* #(or (space? %) (comma? %)) zloc))
+(defn- skip-whitespace-and-commas
+  ([zloc] (skip-whitespace-and-commas zloc z/next*))
+  ([zloc f] (z/skip f (some-fn space? comma?) zloc)))
 
 (defn- skip-clojure-whitespace
   ([zloc] (skip-clojure-whitespace zloc z/next*))
@@ -370,6 +371,7 @@
   {:alias-map                             {}
    :align-binding-columns?                false
    :align-map-columns?                    false
+   :align-single-column-lines?            false
    :aligned-forms                         default-aligned-forms
    :blank-line-forms                      blank-line-forms
    :extra-aligned-forms                   {}
@@ -595,12 +597,9 @@
 (defn sort-ns-references [form]
   (transform form edit-all ns-reference? sort-arguments))
 
-(defn- skip-to-linebreak-or-element [zloc]
-  (z/skip z/right* (some-fn space? comma?) zloc))
-
 (defn- reduce-columns [zloc f init]
   (loop [zloc zloc, col 0, acc init]
-    (if-some [zloc (skip-to-linebreak-or-element zloc)]
+    (if-some [zloc (skip-whitespace-and-commas zloc z/right*)]
       (if (line-break? zloc)
         (recur (z/right* zloc) 0 acc)
         (recur (z/right* zloc) (inc col) (f zloc col acc)))
@@ -623,10 +622,16 @@
                      (map count))
                max 0 (str/split lines #"\r?\n"))))
 
-(defn- max-column-end-position [zloc col]
+(defn- single-column-line? [zloc]
+  (and (line-break? (skip-whitespace-and-commas (z/right* zloc) z/right*))
+       (line-break? (skip-whitespace-and-commas (z/left* zloc) z/left*))))
+
+(defn- max-column-end-position [zloc col align-single-column-lines?]
   (reduce-columns zloc
                   (fn [zloc c max-pos]
-                    (if (= c col)
+                    (if (and (= c col)
+                             (or align-single-column-lines?
+                                 (not (single-column-line? zloc))))
                       (max max-pos (node-end-position zloc))
                       max-pos))
                   0))
@@ -665,7 +670,7 @@
 
 (defn- edit-column [zloc column f]
   (loop [zloc zloc, col 0]
-    (if-some [zloc (skip-to-linebreak-or-element zloc)]
+    (if-some [zloc (skip-whitespace-and-commas zloc z/right*)]
       (let [zloc (if (and (= col column) (not (line-break? zloc)))
                    (f zloc)
                    zloc)
@@ -675,17 +680,24 @@
           zloc))
       zloc)))
 
-(defn- align-one-column [zloc col]
+(defn- align-one-column [zloc col align-single-column-lines?]
   (if-some [zloc (z/down zloc)]
-    (let [start-position (inc (max-column-end-position zloc (dec col)))]
+    (let [start-position (inc (max-column-end-position
+                               zloc (dec col) align-single-column-lines?))]
       (z/up (edit-column zloc col #(pad-to-position % start-position))))
     zloc))
 
-(defn- align-columns [zloc]
-  (reduce align-one-column zloc (-> zloc z/down count-columns range rest)))
+(defn- align-columns [zloc align-single-column-lines?]
+  (reduce #(align-one-column %1 %2 align-single-column-lines?)
+          zloc
+          (-> zloc z/down count-columns range rest)))
 
-(defn align-map-columns [form]
-  (transform form edit-all z/map? align-columns))
+(defn align-map-columns
+  ([form]
+   (align-map-columns form default-options))
+  ([form {:keys [align-single-column-lines?]}]
+   (let [align #(align-columns % align-single-column-lines?)]
+     (transform form edit-all z/map? align))))
 
 (defn- matching-form-index? [zloc [k indexes] context]
   (if (= :all indexes)
@@ -699,16 +711,22 @@
   (and (or (z/list? zloc) (z/list? (z/up zloc)))
        (some #(matching-form-index? zloc % context) form-indexes)))
 
-(defn align-form-columns [form aligned-forms alias-map]
-  (let [ns-name  (find-namespace (z/of-node form))
-        context  {:alias-map alias-map, :ns-name ns-name}
-        aligned? #(matching-form? % aligned-forms context)]
-    (transform form edit-all aligned? align-columns)))
+(defn align-form-columns
+  ([form aligned-forms alias-map]
+   (align-form-columns form aligned-forms alias-map default-options))
+  ([form aligned-forms alias-map {:keys [align-single-column-lines?]}]
+   (let [ns-name  (find-namespace (z/of-node form))
+         context  {:alias-map alias-map, :ns-name ns-name}
+         aligned? #(matching-form? % aligned-forms context)
+         align    #(align-columns % align-single-column-lines?)]
+     (transform form edit-all aligned? align))))
 
 (defn realign-form
   "Realign a rewrite-clj form such that the columns line up into columns."
-  [form]
-  (-> form z/of-node align-columns z/root))
+  ([form]
+   (realign-form form default-options))
+  ([form {:keys [align-single-column-lines?]}]
+   (-> form z/of-node (align-columns align-single-column-lines?) z/root)))
 
 (defn- unalign-from-space [zloc]
   (pad-node (z/right* zloc) (- 1 (node-str-length zloc))))
@@ -811,9 +829,9 @@
          (cond-> (:indentation? opts)
            (reindent indents alias-map opts))
          (cond-> (:align-map-columns? opts)
-           align-map-columns)
+           (align-map-columns opts))
          (cond-> (:align-form-columns? opts)
-           (align-form-columns aligned alias-map))
+           (align-form-columns aligned alias-map opts))
          (cond-> (:remove-trailing-whitespace? opts)
            remove-trailing-whitespace)
          (cond-> (:remove-blank-lines-in-forms? opts)
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -2486,6 +2486,169 @@
          {:align-form-columns? true
           :aligned-forms {'foobar #{0}}}))))
 
+(deftest test-align-form-columns-with-comments
+  (testing "comments at end of bindings should not affect alignment"
+    (is (reformats-to?
+         ["(let [x 1   ; a comment"
+          "      longer 2]"
+          "  (+ x longer))"]
+         ["(let [x      1   ; a comment"
+          "      longer 2]"
+          "  (+ x longer))"]
+         {:align-form-columns? true})))
+  (testing "comments between bindings should not affect alignment"
+    (is (reformats-to?
+         ["(let [binding1 \"value1\""
+          "      ;; This is a comment"
+          "      binding2 \"value2\""
+          "      ;; Another comment"
+          "      longer-binding \"value3\"]"
+          "  (+ binding1 binding2))"]
+         ["(let [binding1       \"value1\""
+          "      ;; This is a comment"
+          "      binding2       \"value2\""
+          "      ;; Another comment"
+          "      longer-binding \"value3\"]"
+          "  (+ binding1 binding2))"]
+         {:align-form-columns? true}))
+    (is (reformats-to?
+         ["(let [a 1"
+          "      ;; comment"
+          "      bb 2"
+          "      ccc 3]"
+          "  (+ a bb ccc))"]
+         ["(let [a   1"
+          "      ;; comment"
+          "      bb  2"
+          "      ccc 3]"
+          "  (+ a bb ccc))"]
+         {:align-form-columns? true})))
+  (testing "let starting with long comment should not affect alignment"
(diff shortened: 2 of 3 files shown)


weavejester:
The solution outlined in #384 should make this patch a little simpler. Because the first column doesn't need alignment, I believe all we need to do is to ignore any single-column lines when calculating the max end position.

I also don't usually say this, but I don't think we need nearly as many test cases. For a fix like this, around 6 is generally enough. 49 is just a little excessive.

jramosg:
@weavejester I hope this fits what you had in mind:[https://github.com/jramosg/cljfmt/commit/69f0bab9c8a6061a64e34cbc9ae4c9dbfdcc249b](https://github.com/jramosg/cljfmt/commit/69f0bab9c8a6061a64e34cbc9ae4c9dbfdcc249b)

weavejester:
Better, but there's a lot of code that I *think* is unnecessary.

What I think we want is this:

```clojure
  (and (line-break? (skip-to-linebreak-or-element zloc))
       (line-break? (z/skip z/left* (some-fn space? comma?) zloc))))

(defn- max-column-end-position [zloc col align-single-column-lines?]
  (reduce-columns zloc
                  (fn [zloc c max-pos]
                    (if (and (= c col)
                             (or align-single-column-lines?
                                 (not (single-column-line? zloc)))
                      (max max-pos (node-end-position zloc))
                      max-pos))
                  0))
```

So we skip any single column line if `align-single-column-lines?` is true.

I might be wrong, but I think that's all we need.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -699,16 +724,19 @@
...
+        context {:alias-map alias-map, :ns-name ns-name}
         aligned? #(matching-form? % aligned-forms context)]
-    (transform form edit-all aligned? align-columns)))
+    (transform form edit-all aligned? #(align-columns % align-single-column-lines?))))
 
 (defn realign-form
   "Realign a rewrite-clj form such that the columns line up into columns."
-  [form]
-  (-> form z/of-node align-columns z/root))
+  [form & [{:keys [align-single-column-lines?]}]]
```
weavejester: This style of adding optional arguments should be avoided for two reasons:
1. It's less efficient than explicit arities
2. It doesn't error if too many arguments are given to the function.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -811,9 +839,9 @@
          (cond-> (:indentation? opts)
            (reindent indents alias-map opts))
          (cond-> (:align-map-columns? opts)
-           align-map-columns)
+           (align-map-columns {:align-single-column-lines? (:align-single-column-lines? opts)}))
```
weavejester: Just pass `opts`. No need to pull the map apart then reassemble it.

jramosg:
> Thanks for the update and your work on this. Could you also make the change detailed in [this comment](https://github.com/weavejester/cljfmt/pull/383#issuecomment-3643173859)? I think we can make the code simpler and clearer by using a `single-column-line?` predicate. There's also no need to factor out the function into `column-end-maximizer` until we need to use that code twice.
> 
> Also please ensure that lines are 80 character or fewer, as detailed in the style guide linked in the contributing guidelines.

# Response to @weavejester's Suggestion

Hello @weavejester ,

Thank you for the suggestion! However, if I use that implementation, some tests fail:

```clojure
Testing cljfmt.core-test

FAIL in (test-align-single-column-lines-option) (/home/jon/sandbox/apps/cljfmt/cljfmt/test/cljfmt/core_test.cljc:2640)
map with wrapped value - :align-single-column-lines? false (default) compact alignment
expected: ["{:key1 1" " :key2" " (fn [a b c d e]" "   (+ a b c d e))" " :key  3}"]
  actual: ["{:key1            1" " :key2" " (fn [a b c d e]" "   (+ a b c d e))" " :key             3}"]

FAIL in (test-align-single-column-lines-option) (/home/jon/sandbox/apps/cljfmt/cljfmt/test/cljfmt/core_test.cljc:2640)
let with wrapped value - :align-single-column-lines? false (default) compact alignment
expected: ["(let [a 2" "      b" "      (fn [e]" "        (+ e 1 1 1 1 1 1 1 1))" "      c 3]" "  (+ a c))"]
  actual: ["(let [a                        2" "      b" "      (fn [e]" "        (+ e 1 1 1 1 1 1 1 1))" "      c                        3]" "  (+ a c))"]

FAIL in (test-align-single-column-lines-option) (/home/jon/sandbox/apps/cljfmt/cljfmt/test/cljfmt/core_test.cljc:2640)
multiple wrapped values still avoid excessive padding with false (default)
expected: ["{:a 1" " :wrapped1" " (fn [x] x)" " :b 2" " :wrapped2" " (fn [y] y)" " :c 3}"]
  actual: ["{:a         1" " :wrapped1" " (fn [x] x)" " :b         2" " :wrapped2" " (fn [y] y)" " :c         3}"]

Ran 27 tests containing 366 assertions.
3 failures, 0 errors.
```

## The Issue

The problem is with how `single-column-line?` detects elements with wrapped values:

```clojure
  (and (line-break? (skip-to-linebreak-or-element zloc))
       (line-break? (z/skip z/left* (some-fn space? comma?) zloc))))
```

```clojure
{:key1 1
 :key2          ; ← Should be detected as single-column
 (fn [x] x)}
```

The problem is that **`z/skip` only moves if the current position matches the predicate**. Since `:key2` is a token (not a space or comma), both `(skip-to-linebreak-or-element zloc)` and `(z/skip z/left* (some-fn space? comma?) zloc)` return `:key2` unchanged without moving in either direction. Then `(line-break? :key2)` returns `false` for both checks, causing the entire `and` expression to return `false`. So it incorrectly fails to detect that `:key2` is on a single-column line (its value is on the next line).

The correct bidirectional implementation would be:

```clojure
  (and (when-some [next-elem (skip-to-linebreak-or-element (z/right* zloc))]
         (line-break? next-elem))
       (when-some [prev-elem (z/skip z/left* (some-fn space? comma?) (z/left* zloc))]
         (line-break? prev-elem))))
```

This moves first with `(z/right* zloc)` and `(z/left* zloc)`, then skips spaces/commas. However, this is more complex than necessary.

## The Solution

The fix is to check **only the right side** - we need to know if the next element after the current one is a line break:

```clojure
  (when-some [next-elem (skip-to-linebreak-or-element (z/right* zloc))]
    (line-break? next-elem)))
```

**Why does the right check alone work?**

Consider these examples:

```clojure
;; Case 1: Value on same line
{:key1 1  ; ← :key1 has value on same line
 ...}

;; Case 2: Value on next line
{:key2    ; ← :key2 alone on this line
 (fn [x] x)  ; ← value on next line
 ...}

;; Case 3: Multiple keys/bindings on same line
{:key1 1 :key2
 (fn [x] x)
 :key3 3}

(let [a 1 b
      (fn [x] x)
      c 3]
  ...)
```

For Case 1 (`:key1`):
- Right check: next element is `1` (not a line break) → **false**
- Result: NOT a single-column line ✓

For Case 2 (`:key2`):
- Right check: next element is newline → **true**
- Result: IS a single-column line ✓

For Case 3 (`:key2` after `:key1 1`, and `b` after `a 1`):
- Right check: next element is newline → **true**
- Result: IS a single-column line ✓
- The alignment is skipped for these because they're single-column lines
- The `wrapped?` check ensures we don't include the wrapped `(fn ...)` in column calculations

The right-side check correctly identifies all cases where a binding/key has its value on the same line versus wrapped to the next line. The `wrapped?` check in `max-column-end-position` handles elements that start on a new line, preventing them from being included in column calculations.

While you could check both left and right with:
You must call `(z/left* zloc)` **first**, then apply `z/skip` with `z/left*`. This is because `z/skip` only continues moving if the current position matches the predicate.
```clojure
  (and (when-some [next-elem (skip-to-linebreak-or-element (z/right* zloc))]
         (line-break? next-elem))
       (when-some [prev-elem (z/skip z/left* (some-fn space? comma?) (z/left* zloc))]
         (line-break? prev-elem))))
```

This more strict check (requiring line breaks on both sides) also passes all tests. However, it's unnecessary complexity since the right-only check, combined with the `wrapped?` check in `max-column-end-position`, handles all the test cases correctly.

I also need to maintain the `(not (wrapped?))` check to skip wrapped values in column calculations:

```clojure
(defn- max-column-end-position [zloc col align-single-column-lines?]
  (reduce-columns zloc
                  (fn [zloc c max-pos]
                    (letfn [(wrapped? []
                              (and (pos? c) (preceded-by-linebreak? zloc)))]
                      (if (and (= c col)
                               (not (wrapped?))  ; ← Skip wrapped values
                               (or align-single-column-lines?
                                   (not (single-column-line? zloc))))
                        (max max-pos (node-end-position zloc))
                        max-pos)))
                  0))
```

Without the `wrapped?` check in `max-column-end-position`, when calculating alignment for column 0 (the keys), it would incorrectly include the wrapped function `(fn ...)` which starts on a new line but is actually a value in column 1.

This produces the expected compact alignment:

```clojure
{:key1 1
 :key2
 (fn [a b c d e]
   (+ a b c d e))
 :key  3}
```

jramosg:
Hello @weavejester 

Thank you for the suggestion! However, when I tested that implementation, some tests failed:

```clojure
Testing cljfmt.core-test

FAIL in (test-align-single-column-lines-option)
map with wrapped value - :align-single-column-lines? false (default) compact alignment
expected: ["{:key1 1" " :key2" " (fn [a b c d e]" "   (+ a b c d e))" " :key  3}"]
  actual: ["{:key1            1" " :key2" " (fn [a b c d e]" "   (+ a b c d e))" " :key             3}"]

FAIL in (test-align-single-column-lines-option)
let with wrapped value - :align-single-column-lines? false (default) compact alignment
expected: ["(let [a 2" "      b" "      (fn [e]" "        (+ e 1 1 1 1 1 1 1 1))" "      c 3]" "  (+ a c))"]
  actual: ["(let [a                        2" "      b" "      (fn [e]" "        (+ e 1 1 1 1 1 1 1 1))" "      c                        3]" "  (+ a c))"]

FAIL in (test-align-single-column-lines-option)
multiple wrapped values still avoid excessive padding with false (default)
expected: ["{:a 1" " :wrapped1" " (fn [x] x)" " :b 2" " :wrapped2" " (fn [y] y)" " :c 3}"]
  actual: ["{:a         1" " :wrapped1" " (fn [x] x)" " :b         2" " :wrapped2" " (fn [y] y)" " :c         3}"]

Ran 27 tests containing 366 assertions.
3 failures, 0 errors.
```

## The Issue

The problem is with how the suggested `single-column-line?` detects elements with wrapped values:

```clojure
  (and (line-break? (skip-to-linebreak-or-element zloc))
       (line-break? (z/skip z/left* (some-fn space? comma?) zloc))))
```

Given this example:

```clojure
{:key1 1
 :key2          ; <- Should be detected as single-column
 (fn [x] x)}
```

The problem is that **`z/skip` only moves if the current position matches the predicate**. Since `:key2` is a token (not a space or comma), both `(skip-to-linebreak-or-element zloc)` and `(z/skip z/left* (some-fn space? comma?) zloc)` return `:key2` unchanged without moving in either direction. Then `(line-break? :key2)` returns `false` for both checks, causing the entire `and` expression to return `false`.

## The Solution

The fix is to **move first, then skip**. We need to call `(z/right* zloc)` and `(z/left* zloc)` before applying the skip operations:

```clojure
  (and (some->> (z/right* zloc)
                skip-to-linebreak-or-element
                line-break?)
       (some->> (z/left* zloc)
                (z/skip z/left* (some-fn space? comma?))
                line-break?)))
```

**Key insight:** The navigation pattern must be:
- **Right side:** Move right with `z/right*`, then skip over spaces/commas, then check for line break
- **Left side:** Move left with `z/left*`, then skip over spaces/commas going left, then check for line break

This bidirectional check ensures the element is truly isolated on its own line (surrounded by line breaks on both sides). Using `some->>` provides clean nil handling - if any step returns nil, the entire expression short-circuits to nil (which is falsy).

With this implementation, `max-column-end-position` can be simplified:

```clojure
(defn- max-column-end-position [zloc col align-single-column-lines?]
  (reduce-columns zloc
                  (fn [zloc c max-pos]
                    (if (and (= c col)
                             (or align-single-column-lines?
                                 (not (single-column-line? zloc))))
                      (max max-pos (node-end-position zloc))
                      max-pos))
                  0))
```

cljfmt/src/cljfmt/core.cljc
```diff
@@ -675,17 +684,30 @@
...
+                               down-zloc
+                               (dec col)
+                               align-single-column-lines?))]
+      (z/up (edit-column down-zloc col #(pad-to-position % start-position))))
     zloc))
 
-(defn- align-columns [zloc]
-  (reduce align-one-column zloc (-> zloc z/down count-columns range rest)))
+(defn- align-columns [zloc align-single-column-lines?]
+  (if-some [down-zloc (z/down zloc)]
```
weavejester: What's the purpose of this check?

cljfmt/src/cljfmt/core.cljc
```diff
@@ -699,16 +721,32 @@
...
+  ([form aligned-forms alias-map]
+   (align-form-columns form aligned-forms alias-map default-options))
+  ([form aligned-forms alias-map {:keys [align-single-column-lines?]}]
+   (let [ns-name (find-namespace (z/of-node form))
+         context {:alias-map alias-map, :ns-name ns-name}
+         aligned? #(matching-form? % aligned-forms context)]
+     (transform form
+                edit-all
+                aligned?
+                #(align-columns % align-single-column-lines?)))))
```
weavejester: More concisely, we could write:

```clojure
(let [ns-name  (find-namespace (z/of-node form))
      context  {:alias-map alias-map, :ns-name ns-name}
      aligned? #(matching-form? % aligned-forms context)
      align    #(align-columns % align-single-column-lines?)]
  (transform form edit-all aligned? align))
```

cljfmt/src/cljfmt/core.cljc
```diff
@@ -699,16 +721,32 @@
...
 (defn realign-form
-  "Realign a rewrite-clj form such that the columns line up into columns."
-  [form]
-  (-> form z/of-node align-columns z/root))
+  "Realign a rewrite-clj form such that the columns line up into columns.
+
+  Options:
+  - :align-single-column-lines? - When true, includes wrapped values in
+    alignment; when false (default), excludes them to prevent excessive
+    padding."
```
weavejester: I'd rather omit the option docs here, as I think it's better to have a central location to document the options. Otherwise we risk two sets of docs getting out of sync.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -699,16 +721,32 @@
...
+  ([form]
+   (realign-form form default-options))
+  ([form {:keys [align-single-column-lines?]}]
+   (-> form
+       z/of-node
+       (align-columns align-single-column-lines?)
+       z/root)))
```
weavejester: This could just be on one like:

```clojure
   (-> form z/of-node (align-columns align-single-column-lines?) z/root)
```

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -2402,6 +2402,13 @@
          {:align-map-columns? true}))))
 
 (deftest test-align-form-columns
+  (testing "empty bindings"
+    (is (reformats-to?
+         ["(let []"
+          "  nil)"]
+         ["(let []"
+          "  nil)"]
+         {:align-form-columns? true})))
```
weavejester: This test doesn't seem to be part of the feature.

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -2486,6 +2493,262 @@
...
+          "      ccc 3]"
+          "  (+ a bb ccc))"]
+         ["(let [;; This is a very long comment at the beginning of the let"
+          "      a   1"
+          "      bb  2"
+          "      ccc 3]"
+          "  (+ a bb ccc))"]
+         {:align-form-columns? true}))))
+
+(deftest test-align-form-columns-edge-cases
```
weavejester: I don't understand how this set of tests is related to the feature the PR is implementing.

As a general rule, the tests added in this PR should fail on master, otherwise they're not testing the feature. This isn't a hard rule - there are exceptions - but it's a decent rule of thumb,

jramosg:
@weavejester done! thank you for the review!

weavejester:
Unfortunately changing the information on the commit is difficult now that it's been merged. I know from experience that this disrupts GitHub's link between the PR and merge commit, and will obviously cause issues with anyone pulling from the master branch. Since this PR was merged 3 weeks ago, I think it's too late to alter.

We can certainly update the PR itself, however, to record correct attribution, either by editing the description or adding a comment as you just did. Presumably you've fixed the contributor information in your other PRs?

PR #389, opened 2026-01-19
Fix case-sensitive sorting in sort-ns-references

### LINKED ISSUE #388: sort-ns-references case-sensitive sorting issue
# Description

The `:sort-ns-references?` option in cljfmt performs case-sensitive string sorting of namespace references, which may not match user expectations for alphabetical ordering.

# Current Behavior

With `:sort-ns-references? true`, namespace references are sorted using case-sensitive string comparison. For example:

```clojure
(ns example
  (:require
   [com.example.answers :as answers]
   [com.example.ERP.service :as erp]))
```

Becomes:

```clojure
(ns example
  (:require
   [com.example.ERP.service :as erp]  ;; 'E' < 'a'
   [com.example.answers :as answers]))
```

# Expected Behavior

Users typically expect case-insensitive alphabetical sorting (like in dictionaries), where "answers" would come before "ERP".

```clojure
(ns example
  (:require
   [com.example.answers :as answers]  ;; 'a' before 'e' (case-insensitive)
   [com.example.ERP.service :as erp]))
```

## Proposed Fix

Modify the sorting to be case-insensitive by converting the sort string to lowercase before comparison.

jramosg:
The :sort-ns-references? option was performing case-sensitive string sorting, causing uppercase namespace references to sort before lowercase ones. This resulted in unexpected ordering like "ERP" before "answers". Modified the node-sort-string function to perform case-insensitive sorting by converting the sort string to lowercase before comparison. Added a test case to verify correct case-insensitive sorting behavior.

Fixes #388

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -589,7 +589,8 @@
       (remove-node-metadata)
       (nodes-string)
       (str/replace #"[\[\]\(\)\{\}]" "")
-      (str/trim)))
+      (str/trim)
+      (str/lower-case)))
 
 (defn sort-arguments [zloc]
   (update-children zloc #(sort-node-arguments-by node-sort-string %)))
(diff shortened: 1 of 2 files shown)


jramosg:
done! thanks @weavejester  !

Issue #391, opened 2026-01-20
Code Execution in Config Files


jramosg:
# Code Execution in Config Files

## Description

cljfmt uses `clojure.core/read-string` to parse `.clj` configuration files, which can execute arbitrary code if the file contains reader macros like `#=`. This poses a significant security risk if configuration files are sourced from untrusted locations or modified by attackers, as malicious code can be executed during the parsing phase.

The vulnerability allows attackers to inject executable code into config files, potentially leading to:
- Code execution with the privileges of the cljfmt process
- Data exfiltration
- System compromise
- Denial of service attacks

This affects all versions of cljfmt that use `read-string` for config parsing.

An attacker could create a malicious config file that executes harmful code:

```clojure
#=(println "Malicious code executed!")
#=(System/exit 1)  ; Example: force JVM exit
{}
```

If this file is named `.cljfmt.clj` and placed in a project directory, running `cljfmt` would execute the code, potentially leading to data exfiltration, system compromise, or denial of service.

### Expected Behavior

Configuration files should be treated as data only, without executing any code. Only safe, non-executable parsing should be used.

### Reproduction Steps

1. Create a file named `.cljfmt.clj` with content: `#=(println "Code executed!") {}`
2. Run `cljfmt fix <some-file.clj>`
3. Observe the print output, confirming code execution.

This should throw an exception instead of executing.

PR #392, opened 2026-01-20
Fix security vulnerability: prevent code execution in config files

jramosg:
Fix security vulnerability: prevent code execution in config files
Replace read-string with edn/read-string for .clj config files to prevent
arbitrary code execution via reader macros like #=. This maintains support
for safe reader macros like #re while eliminating the security risk.

Add tests to verify that malicious reader macros throw exceptions and
valid configs still work.

Fixes https://github.com/weavejester/cljfmt/issues/391

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,10 @@ files in the current and parent directories:
 
 The configuration file should contain a map of options.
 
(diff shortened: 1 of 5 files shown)


jramosg:
changed, thanks!

PR #393, opened 2026-01-22
Add refer-map support for custom indentation rules

### LINKED ISSUE #368: Format vars that are introduced with :refer
I am looking at adopting cljfmt. One thing I noticed is that it seems like vars that are introduced via :refer are not aliased to their namespaced cljfmt config. E.g.

```clj
(ns my.ns
  (:require [my.util :refer [timed]))

(timed "x"
  :a
  :b)
```

If I have an indent rule for `my.util/timed` in my `cljfmt.edn`, `timed` isn't indented. However, if I change to

```clj
(ns my.ns
  (:require [my.util :as util))

(util/timed "x"
  :a
  :b)
```

Then it will be indented correctly. Is it possible/should it be possible for vars introduced with :refer to use the namespaced indent config?

jramosg:
This change introduces a `:refer-map` option that allows users to define mappings of referred symbol names to their fully qualified namespace names. This is particularly useful for applying indentation rules to unqualified symbols that need to match against qualified indent rules, enhancing the flexibility of the formatting tool.

Fixes #368

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -302,11 +302,17 @@
   (when-let [ns-str (namespace possible-sym)]
     (symbol (get alias-map ns-str ns-str) (name possible-sym))))
 
+(defn- qualify-symbol-by-refer-map [possible-sym refer-map]
+  (when-not (namespace possible-sym)
+    (when-let [ns-str (get refer-map (str possible-sym))]
+      (symbol ns-str (name possible-sym)))))
+
 (defn- qualify-symbol-by-ns-name [sym ns-name]
   (when ns-name (symbol (name ns-name) (name sym))))
 
 (defn- fully-qualified-symbol [sym context]
   (or (qualify-symbol-by-alias-map sym (:alias-map context))
+      (qualify-symbol-by-refer-map sym (:refer-map context))
       (qualify-symbol-by-ns-name sym (:ns-name context))))
 
 (defn- string-matches-key-part? [s key-part]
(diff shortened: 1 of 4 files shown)


cljfmt/src/cljfmt/core.cljc
```diff
@@ -838,7 +871,12 @@
                           (:extra-blank-line-forms opts))
          alias-map #?(:clj  (merge (alias-map-for-form form)
                                    (stringify-map (:alias-map opts)))
-                      :cljs (stringify-map (:alias-map opts)))]
+                      :cljs (stringify-map (:alias-map opts)))
+         refer-map #?(:clj  (merge (refer-map-for-form form)
+                                   (stringify-map (:refer-map opts)))
+                      :cljs (stringify-map (:refer-map opts)))
+
```
weavejester: This blank line can be omitted.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -799,19 +806,45 @@
...
+     (defn- parent-namespace [grandparent-node]
+       (when-not (ns-require-form? grandparent-node)
+         (when (or (z/vector? grandparent-node)
+                   (z/list? grandparent-node))
+           (first (z/child-sexprs grandparent-node)))))
+
+     (defn- ns-str [parent-namespace current-ns]
+       (if parent-namespace
+         (format "%s.%s" parent-namespace current-ns)
+         (str current-ns)))
```
weavejester: Maybe `join-ns-str` would be better, to indicate that this is for joining two ns strings together.

weavejester (changes requested):
Apologies for the delay in reviewing this. Overall great work - I just have some small suggestions about naming functions.

weavejester:
Merged! I'll cut a release soon with this added.

PR #394, opened 2026-01-22
Add support to normalize newlines at file end

### LINKED ISSUE #358: Automatically remove multiple blank lines at end of file
It would be nice to automatically format a file like 

```clj
(ns my.namespace)

(whatever ...)


```

to 

```clj
(ns my.namespace)

(whatever ...)
```

(make sure the file ends in exactly one newline character)

jramosg:
This change introduces an option to insert a final newline at the end of files when formatting. It includes logic to handle various edge cases, such as preserving existing newlines and removing extra newlines or trailing whitespace. Tests have been added to ensure correct behavior when the option is enabled or disabled. 
Fixes #358

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -381,6 +381,7 @@
    :indent-line-comments?                 false
    :indentation?                          true
    :indents                               default-indents
+   :normalize-newlines-at-file-end?       false
    :insert-missing-whitespace?            true
    :remove-blank-lines-in-forms?          false
    :remove-consecutive-blank-lines?       true
@@ -501,6 +502,10 @@
 
 (defn remove-trailing-whitespace [form]
   (transform form edit-all trailing-whitespace? z/remove*))
+
+(defn normalize-newlines-at-file-end [s]
+  (cond-> (str/trimr s)
+    (not (str/blank? s)) (str "\n")))
 
 (defn- replace-with-one-space [zloc]
   (z/replace* zloc (whitespace 1)))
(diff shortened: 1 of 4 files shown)


cljfmt/src/cljfmt/core.cljc
```diff
@@ -502,6 +503,24 @@
...
+
+(defn insert-final-newline [form]
+  (let [form-no-trailing (transform form edit-all trailing-newline? z/remove*)
+        empty-or-whitespace? (or (empty? (n/children form-no-trailing))
+                                 (only-whitespace? form-no-trailing))]
+    (if empty-or-whitespace?
+      form-no-trailing
+      (n/replace-children form-no-trailing
+                          (concat (n/children form-no-trailing)
+                                  [(n/newlines 1)])))))
```
weavejester: Do we need to do this via `reformat-form`? This would be a lot simpler if we just added it to `reformat-string` instead:

```clojure
(cond-> (:normalize-newlines-at-file-end options)
  (-> str/trimr (str "\n")))
```

weavejester (changes requested):

cljfmt/src/cljfmt/core.cljc
jramosg: you're right, I did some small modifications to handle edge cases like:
```clojure
(reformats-to?
           [""]
           [""]
           {:normalize-newlines-at-file-end true})
```
```clojure
(reformats-to?
           [""
            " "
            ""]
           [""]
           {:normalize-newlines-at-file-end true})
```
We must check that string is not empty before we add /n at the end

```clojure
(defn normalize-newlines-at-file-end [form]
  (let [s (s/trimr (n/string form))]
    (cond-> s (seq s) (str "\n"))))
```

PR #410, opened 2026-04-01
Fix the config key for `blank-lines-separate-alignment` in README

beatriz:
All other configuration keys were present as the full keyword, except for `blank-lines-separate-alignment`. This fixes it, as I was pretty confused why the config wasn't working when I had copied it from the README as all others 😄


beatriz:
Yes, sorry about that! I fixed the commit message.

PR #411, opened 2026-04-01
Fix `single-column-line?` behavior for last elements

beatriz:
I noticed that the `:single-column-line?` rule is not currently respected when the single-column element is the last one in a map. This PR fixes that.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -631,7 +631,8 @@
                max 0 (str/split lines #"\r?\n"))))
 
 (defn- single-column-line? [zloc]
-  (and (line-break? (skip-whitespace-and-commas (z/right* zloc) z/right*))
+  (and (let [zloc (skip-whitespace-and-commas (z/right* zloc) z/right*)]
+         (or (nil? zloc) (line-break? zloc)))
        (line-break? (skip-whitespace-and-commas (z/left* zloc) z/left*))))
 
 (defn- node-str-length [zloc]
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -2713,6 +2713,19 @@
           " (fn [a b c d e]"
           "   (+ a b c d e))"
           " :key  3}"]
+         {:align-map-columns? true}))
+
+    (is (reformats-to?
+         ["{:key-one 1"
+          " :key3 3"
+          " :key2"
+          " (fn [a b c d e]"
+          "   (+ a b c d e))}"]
+         ["{:key-one 1"
+          " :key3    3"
+          " :key2"
+          " (fn [a b c d e]"
+          "   (+ a b c d e))}"]
          {:align-map-columns? true})))
   (testing "let with wrapped value - :align-single-column-lines? false (default) compact alignment"
     (is (reformats-to?


cljfmt/src/cljfmt/core.cljc
```diff
@@ -631,8 +631,10 @@
                max 0 (str/split lines #"\r?\n"))))
 
 (defn- single-column-line? [zloc]
-  (and (line-break? (skip-whitespace-and-commas (z/right* zloc) z/right*))
-       (line-break? (skip-whitespace-and-commas (z/left* zloc) z/left*))))
+  (let [right* (z/right* zloc)]
+    (and (or (nil? right*)
+             (line-break? (skip-whitespace-and-commas right* z/right*)))
+         (line-break? (skip-whitespace-and-commas (z/left* zloc) z/left*)))))
```
weavejester: Please correct me if I'm wrong, but wouldn't we want to check for nil after skipping the whitespace and commas? i.e.

```clojure
  (and (let [zloc (skip-whitespace-and-commas (z/right* zloc) z/right*)]
         (or (nil? zloc) (line-break? zloc)))
       (line-break? (skip-whitespace-and-commas (z/left* zloc) z/left*)))
```

Otherwise a single column line ending in a comma or whitespace wouldn't trigger this rule.

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -2748,6 +2713,19 @@
...
+          " :key3 3"
+          " :key2"
+          " (fn [a b c d e]"
+          "   (+ a b c d e))}"]
+         ["{:key1 1"
+          " :key3 3"
+          " :key2"
+          " (fn [a b c d e]"
+          "   (+ a b c d e))}"]
          {:align-map-columns? true})))
```
weavejester: Ideally we should have some invalid indentation so that there's a change in the two expression strings, otherwise we can't distinguish between it working and it doing nothing. So perhaps:

```clojure
    (is (reformats-to?
         ["{:key-first 1"
          " :key2 3"
          " :key3"
          " (fn [a b c d e]"
          "   (+ a b c d e))}"]
         ["{:key-first 1"
          " :key2      3"
          " :key3"
          " (fn [a b c d e]"
          "   (+ a b c d e))}"]
         {:align-map-columns? true})))
```

cljfmt/src/cljfmt/core.cljc
beatriz: Yes, you are correct! Fixed it.

cljfmt/test/cljfmt/core_test.cljc
beatriz: Makes total sense, I've adjusted the test.

cljfmt/src/cljfmt/core.cljc
weavejester: Thanks! Could you use the code I suggested in my previous comment? It makes it a little more concise and constrains the scope of the `let` clause.

beatriz: Of course!

beatriz: This is ready to review, when you have the chance 🙂

weavejester:
This all looks good! Can you squash down your commits and give them an appropriate commit message like:

```
Fix single-column-line? behavior for last elements
```

beatriz:
Of course, done!

weavejester:
Merged. Thanks!

beatriz:
Cool! Do you have a release planned for soon?

weavejester:
Just doing it right now :)

Issue #114, opened 2018-02-14
Request: Format enforce 80 column layout


cneira:
Is it possible to have an option to format code to 80 columns?

daveyarwood:
I also think this would be a good addition.

For me, the motivation is that vim-fireplace, which is sort of "the" Clojure REPL integration plugin for Vim, breaks the standard Vim keybinding `gq` which is used to format code such that each line is limited to 80 characters, and instead uses cljfmt to format code. So, it would be nice if it were possible to configure cljfmt to format code to 80 columns so that `gq` can behave like we expect in Vim. [This discussion on Clojurians Slack](https://clojurians-log.clojureverse.org/vim-fireplace/2017-06-07.html) brought me here.

I think ideally, the cljfmt option would allow you to specify a text-width parameter (a number, not necessarily 80) which, when present, formats code such that each line is limited to that number of columns; and when not present, no width-related formatting is done.

Then we could set vim-fireplace up to pass through the user's vim `textwidth` setting through as that parameter and `gq` ought to work seamlessly.

weavejester:
I'm fine with this, and it adheres to the Clojure Style Guide. However, I won't have a lot of time to spare to spend on this; I'll review and merge PRs, but I won't be working on cljfmt myself, for a while.

rplevy:
Just to put it out there, if people are coming across this issue looking for an 80-char Clojure source code formatter, that is one of the features of https://github.com/kkinnear/zprint

carlosgrinan:
Has this been implemented yet? I don´t see it on https://github.com/weavejester/cljfmt?tab=readme-ov-file#configuration

weavejester:
No, this hasn't been implemented. I'll reopen the issue.

Issue #413, opened 2026-04-10
Enforce where line breaks are allowed


weavejester:
Some forms have conventions on where line breaks are permitted. For example:

```clojure
;; ok
(if x y z)

;; ok
(if x
  y
  z)

;; bad
(if x y
  z)
```

For more complex forms, like `defn`, this might be affected by the type of arguments:

```clojure
;; ok
(defn foo
  "docstring"
  [x y z])

;; less ok
(defn foo "docstring"
  [x y z])
```

Ideally we want some syntax that defines the positions where newlines are valid, and to add newlines that are missing. This issue has been opened to discuss what form that syntax will take.

This issue replaces issue #60, #61, #105, #317 and #412.

kranstrom:
@weavejester I took some time to put together a proposal for this that I hope adheres to your existing project structure/configuration. If at all interested, I can open a PR:

https://github.com/kranstrom/cljfmt/commit/53a63e0183ea7551ae59edc6bd861d21aa1e4d1c

weavejester:
Thank you for putting together that commit, but ideally we should decide on a design before writing any code, otherwise you might write a lot of code for a design that's not adopted. Unfortunately, I think that might be the case here, as it differs from the approaches I was considering, and the amount of code and special rules that are needed is rather more than I was expecting.

There is one main issue with any design that solves this issue, which is this: if there are multiple correct ways of formatting, which correct version do we choose?

To illustrate this problem, let's suppose we've decided there are two valid ways of line-breaking an `if` expression:

```clojure
;; 1.
(if a? b c)

;; 2.
(if a?
  b
  c)
```

Now let's suppose we have a badly formatted expression:

```clojure
(if a? b
  c)
```

We can either add a line-break or remove a line break, and either option will get us to a correct result. In this case it's a coin flip, but for a more complex expression we need to consider how many changes are needed to bring the form into a correctly formatted state, and presumably choose the minimum.

I've given this whole issue some hammock time, and my best idea so far is to use a regular expression, in the same way that spec operates:

```clojure
{if [:or (_ _ _)
         (_ \n _ \n _)]}
```

This allows us to spot a form that's out of alignment, and we can scale this to some fairly complex patterns, particularly if we add literal type matching as well:

```clojure
{defn (:sym               ; name
       [:? \n :str \n]    ; docstring
       :vec \n            ; arglist 
       [:? [:* _ \n] _])} ; body
```

[...]

PR #414, opened 2026-04-17
Add optional column-oriented wrapping via :max-line-length

ispandey81:
Refs https://github.com/weavejester/cljfmt/issues/114

Apply :max-line-length as a structural pass: break only between direct children of eligible single-line forms using existing width accounting. Cover lists (including ns :import / :use), vectors, maps, sets, fn literals, namespaced maps, and inner collections in #? / #?@ reader conditionals; skip layout changes inside other reader macros and inside ns :require libspec vectors (row breaks only between those vectors). When :max-line-length is set, gate map keypair newlines on line width; keep unconditional keypair splitting when only
:split-keypairs-over-multiple-lines? is enabled. Document behavior and split/max interaction in README; extend core tests accordingly.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -278,6 +278,12 @@
 (defn- reader-conditional? [zloc]
   (and (reader-macro? zloc) (#{"?" "?@"} (-> zloc z/down token-value str))))
 
+(defn- structural-wrap-under-forbidden-reader-macro? [zloc]
+  (some (fn [a]
+          (and (reader-macro? a)
+               (not (reader-conditional? a))))
+        (take-while some? (iterate z/up* (z/up* zloc)))))
+
 (defn- find-next-keyword [zloc]
   (z/find zloc z/right #(n/keyword-node? (z/node %))))
 
@@ -397,7 +403,8 @@
    :remove-surrounding-whitespace?        true
    :remove-trailing-whitespace?           true
    :sort-ns-references?                   false
-   :split-keypairs-over-multiple-lines?   false})
+   :split-keypairs-over-multiple-lines?   false
+   :max-line-length                       nil})
 
 (defmulti ^:private indenter-fn
   (fn [_sym _context [type & _args]] type))
@@ -483,9 +490,6 @@
 
 (defn- insert-newline-left [zloc]
   (z/insert-left* zloc (n/newlines 1)))
-
-(defn split-keypairs-over-multiple-lines [form]
-  (transform form edit-all map-key-without-line-break? insert-newline-left))
 
 (defn reindent
   ([form]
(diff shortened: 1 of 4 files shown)


ispandey81:
Hi @weavejester , this is my first stab at https://github.com/weavejester/cljfmt/issues/114 
Let me know if this is a step in the right direction.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -630,6 +634,108 @@
...
+          max-on? (pos-int? max-len)]
+      (cond
+        (and split? (not max-on?)) true
+        max-on? (> (node-end-position zloc) max-len)
+        :else false))))
+
+(defn- split-map-keypairs-for-layout [form opts]
+  (transform form edit-all
+             #(map-key-needs-split-for-layout? % opts)
+             insert-newline-left))
```
weavejester, line 650: Can you explain why the `split-keypairs-over-multiple-lines` function needs to be changed in this way? I think there needs to be a very good reason to add complexity here.

cljfmt/src/cljfmt/core.cljc
```diff
@@ -630,6 +634,108 @@
...
+        (and (token? p) (= :require (z/sexpr p))) false
+        :else true))))
+
+(defn- wrap-ns-require-libspec-rows-to-max-length [form max-len]
+  (transform form edit-all
+             (fn [zloc]
+               (and (libspec-vector-not-first-after-require? zloc)
+                    (single-line-direct-children? (z/up zloc))
+                    (> (node-end-position zloc) max-len)))
+             insert-newline-left))
```
weavejester, line 698: Can you explain why there's a separate function for libspecs? Why is breaking the lines here different to other forms?

cljfmt/src/cljfmt/core.cljc
```diff
@@ -661,8 +767,8 @@
     zloc))
 
 (defn- pad-node [zloc padding]
-  (-> (update-space-left zloc padding)
-      (z/subedit-> (pad-inside-node padding))))
+  (z/subedit-node (update-space-left zloc padding)
+                  #(pad-inside-node % padding)))
```
weavejester, line 771: Why was this change made?

cljfmt/src/cljfmt/core.cljc
```diff
@@ -630,6 +634,108 @@
...
+    (if (pos-int? max-len)
+      (-> form
+          (wrap-ns-require-libspec-rows-to-max-length max-len)
+          (transform edit-all
+                     (fn [zloc]
+                       (and (max-len-structural-wrap-target? zloc)
+                            (single-line-direct-children? zloc)
+                            (not (structural-wrap-under-forbidden-reader-macro? zloc))))
+                     #(wrap-max-line-length-at-child-boundaries % max-len)))
+      form)))
```
weavejester, line 737: I think we should be aiming for a function like:

```clojure
(defn limit-line-length [form {:keys [max-line-length]}]
  (transform edit-all #(needs-line-break? % max-line-length) add-line-break))
```

cljfmt/src/cljfmt/core.cljc
```diff
@@ -630,6 +634,108 @@
...
+        zloc
+        (let [last-zloc
+              (loop [current first-el]
+                (if-some [nxt (-> current z/right* (skip-whitespace-and-commas z/right*))]
+                  (recur (if (> (node-end-position nxt) max-len)
+                           (insert-newline-left nxt)
+                           nxt))
+                  current))]
+          (z/up last-zloc))))))
+
```
weavejester, line 725: This is looks like the logic for line breaks depends on the form, but why is that the case? Why not add line breaks at the first possible place prior to the max-line-length?

If certain forms are going to need specific line break, that sounds like something that should be configurable (see #413).

weavejester:
Thanks for the PR, I have a few comments which I've made inline.

Generally I think this needs to be simpler and more general. Where the line break happens should be very predictable, and shouldn't depend on what the outer form is.

I also note that the PR doesn't seem to include calls to the indentation functions, unless I've missed that somehow? You need to know how a line will be indented in order to know whether a line break is in an appropriate place. For instance:

```clojure
;; too long a line
(aaaaaaaaaaaaaaaaaaaaaaaaaaaa bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb cccccccccccccccccccccccccccccccccccccccccccccccccccccccc)

;; breaking it up here also leads to too long a line due to indent
(aaaaaaaaaaaaaaaaaaaaaaaaaaaa bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
                              cccccccccccccccccccccccccccccccccccccccccccccccccccccccc)


;; so it needs to be split like so:
(aaaaaaaaaaaaaaaaaaaaaaaaaaaa
 bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
 cccccccccccccccccccccccccccccccccccccccccccccccccccccccc)

;; however, if the indentation rules were different, this might be valid:
(aaaaaaaaaaaaaaaaaaaaaaaaaaaa bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
  cccccccccccccccccccccccccccccccccccccccccccccccccccccccc)
```

cljfmt/src/cljfmt/core.cljc
ispandey81: If we keep the old behavior when both are on, the keypair pass always explodes maps onto many lines before the line-length pass runs. That makes :max-line-length almost meaningless for maps (it all stems from my approach being dependent on the outer form which I agree is not generic)

My intent with this change: with a positive max and keypair, breaks on maps are width-gated; split-only (no max) stays exactly the old “always split qualifying keys” behavior.

cljfmt/src/cljfmt/core.cljc
ispandey81: my bad, this should be reset to the way it was.. Initial drafts of my change were reporting an error for this when I ran bb lint

cljfmt/src/cljfmt/core.cljc
ispandey81: I will take this feedback and apply it for my next iteration of this PR

cljfmt/src/cljfmt/core.cljc
ispandey81: General structural wrap inserts newlines between direct children of lists/vectors. Breaking “at the first child boundary” inside [my.ns :as m] splits the same require entry across lines, which is not ideal. If we later get configurable break rules ([#413](https://github.com/weavejester/cljfmt/issues/413)), this could become one configured rule among others.

cljfmt/src/cljfmt/core.cljc
ispandey81: My thinking was hard wrap at max-line-length is going to be unsafe e.g. 
very-long-namespace/foo or keyword-with-dashes must stay intact.
String literals cannot be broken without changing semantics etc

I will think about not being dependent on the form and make this more generic

weavejester: By "the first possible place" I mean the first possible place that doesn't change the semantic meaning.

cljfmt/src/cljfmt/core.cljc
weavejester: I think we want to attempt this feature in two steps:

1. Get a basic line-breaking feature
2. Create some sort of per-form syntax for preferring line break locations

I don't think we should try to hard code specific behavior, as (a) we'll have to eventually remove it in favor of a configurable system, and (b) we don't want to break things when we switch over systems.

weavejester:
I think before any more code is written, we need to define the algorithm we're using to divide up lines. This needs to:

1. Work with lines of arbitrary length
2. Be able to split a line up more than once if necessary
3. Take into account indentation
4. Accept that some lines cannot be split up without user intervention (e.g. long string literals)

Splitting one line into two seems relatively straightforward: we start from the line break and step left until we get a line break position that works. However, splitting one line into three or more seems more complex. Consider a situation where we wrap at 10 characters:

```clojure
;123456789
(aa bb (cc dd ee ff))
```

How should this be wrapped? Presumably we want to try to minimize the number of new lines under most circumstances, but are we trying every permutation, or relying on some greedy matching algorithm that works *most* of the time?

PR #418, opened 2026-05-22
Fix parsing ns requires with reader conditionals

### LINKED ISSUE #417: Reader conditionals in `ns` `:require` are not handled correctly
I was trying to enable the new `:remove-blank-lines-in-forms?` option I added a while back and ran into an issue with it.

Not saying I actually want to allow blank lines in `defprotocol` but this is an example that reproduces the bug:

With a config like this:

```clj
{:remove-blank-lines-in-forms? true
 :extra-blank-line-forms       {potemkin/defprotocol+ :all}
 :extra-indents                {potemkin/defprotocol+ [[:block 1] [:inner 1]]}}
```

The file 

```clj
(ns x
  (:require
   #?@(:clj [[potemkin :as p]])))

#?(:clj
   (p/defprotocol+ Protocol
     (my-method [_this]
       "This is a great method.")

     (another-method [_this]
       "Another nice method.")))
```

Gets reformatted to 

```clj
(ns x
  (:require
   #?@(:clj [[potemkin :as p]])))

#?(:clj
   (p/defprotocol+ Protocol
     (my-method [_this]
       "This is a great method.")
     (another-method [_this]
       "Another nice method.")))
```

(notice the blank line is removed).

If I remove the reader conditional in the require:

```clj
(ns x-test
  (:require
   [potemkin :as p]))
```

 then it formats the file as I'd expect.

So I guess there is probably an issue with `cljfmt.core/alias-map-for-form` not handling reader conditionals correctly, which means this issue probably affects a lot of other features that rely on it as well. 

I'll dig into it and open a PR if I can fix it

camsaul:
Fixes #417

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -820,7 +820,10 @@
           (when-not (ns-require-form? grandparent-node)
             (when (or (z/vector? grandparent-node)
                       (z/list? grandparent-node))
-              (first (z/child-sexprs grandparent-node))))))
+              (some-> (z/find (-> grandparent-node z/down skip-meta)
+                              (comp skip-meta z/right)
+                              symbol-node?)
+                      z/sexpr)))))
 
 #?(:clj (defn- join-ns-str [parent-namespace current-ns]
           (if parent-namespace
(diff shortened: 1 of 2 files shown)


weavejester:
Thanks for the fix, this all looks good so I'll go ahead and merge it in.

Issue #335, opened 2024-02-11
Un-nest namespace functionality


velios:
Tell me if you were ready to accept PR into the code base allowing formatting namespaces in the ways described in the [article by Stuart Sierra how-to-ns](https://stuartsierra.com/2016/clojure-how-to-ns.html)?

I mean this part. Un-nest namespace feature
```clojure
;; Good
(:require
 [com.example.client]
 [com.example.routes]
 [com.example.server])
;; Bad
(:require
 (com.example client server routes))
```

And also this horizontal space saving feature
```clojure
;; Good
(:require
 [com.example.client]
 [com.example.routes]
 [com.example.server])
;; Bad
(:require [com.example.client]
          [com.example.routes]
          [com.example.server])
```

There are separate utilities like [gfredericks/how-to-ns](https://github.com/gfredericks/how-to-ns) for performing this functionality, but it seems it would be appropriate to transfer this to your tool. If you are not interested for code like this in cljfmt codebase, then let me know and I won’t waste time preparing it. I'm not interested in doing this as a separate tool.

weavejester:
That would be fine as two optional settings.

kranstrom:
@weavejester @velios I didn't see any recent activity on this one, so decided to propose a PR for one of the two requested changes in this issue. Thanks!

weavejester:
Support for determining line breaks in the `ns` form has been folded into the more general issue #413.

PR #422, opened 2026-07-15
Add support for line-breaks-in-ns

kranstrom:
Adds new configuration option `:line-breaks-in-ns` that allows for consistent line breaks within `ns` `:require`, `:import`, etc.

The reason for the change is:

- Consistent formatting of `ns`
- Easier manual sorting (e.g. Emacs `sort-lines`)
- Improved readability of `ns` declarations
- Cleaner diffs for dependency changes

The implementation requires an input of:

- `:none` (default) - no enforceable line breaks in `ns`.

- `:multiple` - ensures singular dependencies are moved to the same line as the `ns` reference symbol. Multiple dependencies are broken out into separate lines below the `ns` reference symbol. `:multiple` is consistent with [Clojure Style Guide: Line breaks in ns](https://github.com/bbatsov/clojure-style-guide#line-breaks-in-ns)

- `:always` - ensures all dependencies are moved to separate lines from the `ns` reference symbol. `:always` is consistent with [Stuart Sierra: How to ns - Line Breaks](https://stuartsierra.com/2016/clojure-how-to-ns.html#line-breaks)

This commit resolves one of two features requested by #335.

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,6 +370,19 @@ In order to load the standard configuration file from Leiningen, add the
   See [INDENTS.md][] for a complete explanation. This will **replace**
   the default indents.
 
+* `:line-breaks-in-ns` -
+  - `:none` if cljfmt should leave line breaks within `ns` alone.
+  - `:multiple` to only break lines when there are multiple dependencies,
+    while keeping single dependencies next to the `ns` reference symbol.
+    This follows [Clojure Style Guide: Line breaks in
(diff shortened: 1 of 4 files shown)


weavejester:
Rather than having special syntax for specifying linebreaks in the `ns` form, it makes more sense to have a general syntax for defining where linebreaks are allowed. Issue #413 was opened to discuss this.

kranstrom:
@weavejester understood, I missed this other issue.  I'm closing this one in favor of the outcome from #413.

PR #423, opened 2026-07-24
fix column alignment: single-column first elements, gap-limit anchor …

sirmspencer:
…demotion


sirmspencer:
sorry this is premature

PR #424, opened 2026-07-24
fix single-column-line? to detect first element in collection

sirmspencer:
@weavejester I can't remember if you want an issue per PR, so I included the details here.  I can make it an issue if you prefer.

### LLM notice
I have reviewed all of what is being submitted.  The PR description was partially generated with llm to show the current and expected.  The test cases were partially generated with llm to cover all the permutations. 

## Problem

`:align-single-column-lines? false` does not suppress anchor inflation when the wrapped form is the **first** binding in a `let` or the first entry in a map. The first-position element is always treated as an alignment anchor, producing the same excessive padding as `:align-single-column-lines? true`.

The underlying cause is in `single-column-line?`, which identifies whether a node is the sole element on its line. It checks for a linebreak to the left of the node, but the first element in a collection has no left sibling — `z/left*` returns nil, `line-break?` of nil returns false, and the element is never recognized as single-column regardless of what is on its right.

### Form columns

Config: `{:align-form-columns? true, :align-single-column-lines? false}`

Input:
```clojure
(let [{:keys [endpoint mapper]}
      (get-in config [:auth])
      token (get response "access_token")
      type (get response "token_type")]
  {:token token})
```

Output — first-position wrapped binding used as anchor despite `false`:
```clojure
(let [{:keys [endpoint mapper]}
      (get-in config [:auth])
      token                     (get response "access_token")
      type                      (get response "token_type")]
  {:token token})
```

### Map columns

Config: `{:align-map-columns? true, :align-single-column-lines? false}`

Input:
```clojure
{:very-long-map-key
 wrapped-value
 :host "localhost"
 :port 5432
 :timeout 30}
```

Output — first-position entry used as anchor despite `false`:
```clojure
{:very-long-map-key
 wrapped-value
 :host              "localhost"
 :port              5432
 :timeout           30}
```

### Form columns

```clojure
;; first-position wrapped binding excluded; token and type align locally
(let [{:keys [endpoint mapper]}
      (get-in config [:auth])
      token (get response "access_token")
      type  (get response "token_type")]
  {:token token})
```

### Map columns

```clojure
;; first-position entry excluded; :host/:port align to :timeout
{:very-long-map-key
 wrapped-value
 :host    "localhost"
 :port    5432
 :timeout 30}
```

## Fix

`single-column-line?` currently requires a linebreak to the left:

```clojure
(line-break? (skip-whitespace-and-commas (z/left* zloc) z/left*))
```

A nil result (first element in collection) is treated as false. The fix treats nil
the same as a linebreak:

```clojure
(let [left (skip-whitespace-and-commas (z/left* zloc) z/left*)]
  (or (nil? left) (line-break? left)))
```

Tests cover all three positions (first, middle, last) for both `align-form-columns?`
and `align-map-columns?`, with both `true` and `false` flag values, added to the
existing `test-align-single-column-lines-option` deftest.

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -634,9 +634,10 @@
                max 0 (str/split lines #"\r?\n"))))
 
 (defn- single-column-line? [zloc]
-  (and (let [zloc (skip-whitespace-and-commas (z/right* zloc) z/right*)]
-         (or (nil? zloc) (line-break? zloc)))
-       (line-break? (skip-whitespace-and-commas (z/left* zloc) z/left*))))
+  (and (let [right (skip-whitespace-and-commas (z/right* zloc) z/right*)]
+         (or (nil? right) (line-break? right)))
+       (let [left (skip-whitespace-and-commas (z/left* zloc) z/left*)]
+         (or (nil? left) (line-break? left)))))
 
 (defn- node-str-length [zloc]
   (-> zloc z/node n/string count))
--- a/cljfmt/test/cljfmt/core_test.cljc
+++ b/cljfmt/test/cljfmt/core_test.cljc
@@ -2798,6 +2798,32 @@
           " :key             3}"]
          {:align-map-columns?         true
           :align-single-column-lines? true})))
+  (testing "map with first-position wrapped value, true"
+    (is (reformats-to?
+         ["{:very-long-map-key"
+          " wrapped-value"
+          " :host \"localhost\""
+          " :port 5432"
+          " :timeout 30}"]
+         ["{:very-long-map-key"
+          " wrapped-value"
+          " :host              \"localhost\""
+          " :port              5432"
+          " :timeout           30}"]
+         {:align-map-columns?         true
+          :align-single-column-lines? true})))
+  (testing "map with last-position wrapped value, true"
+    (is (reformats-to?
+         ["{:short 1"
+          " :timeout 30"
+          " :very-long-key"
+          " wrapped}"]
+         ["{:short         1"
+          " :timeout       30"
+          " :very-long-key"
+          " wrapped}"]
+         {:align-map-columns?         true
+          :align-single-column-lines? true})))
   (testing "let with wrapped value - :align-single-column-lines? true causes excessive padding"
     (is (reformats-to?
          ["(let [a 2"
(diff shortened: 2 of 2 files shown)


cljfmt/src/cljfmt/core.cljc
```diff
@@ -634,9 +634,10 @@
                max 0 (str/split lines #"\r?\n"))))
 
 (defn- single-column-line? [zloc]
-  (and (let [zloc (skip-whitespace-and-commas (z/right* zloc) z/right*)]
-         (or (nil? zloc) (line-break? zloc)))
-       (line-break? (skip-whitespace-and-commas (z/left* zloc) z/left*))))
+  (and (let [right (skip-whitespace-and-commas (z/right* zloc) z/right*)]
+         (or (nil? right) (line-break? right)))
+       (let [left (skip-whitespace-and-commas (z/left* zloc) z/left*)]
+         (or (nil? left) (line-break? left)))))
```
weavejester, line 640: How about we use `some->` here?

```clojure
  (and (some-> zloc z/right* (skip-whitespace-and-commas z/right*))
       (some-> zloc z/left*  (skip-whitespace-and-commas z/left*)))
```

weavejester (commented):
Thanks for the PR. Can you change the commit message so it adheres to [the seven rules of a great Git commit message](https://chris.beams.io/git-commit#seven-rules) as per the contributing guidelines. I also have a suggestion regarding the use of `some->` that I've added as a comment.

cljfmt/test/cljfmt/core_test.cljc
```diff
@@ -2798,6 +2798,32 @@
           " :key             3}"]
          {:align-map-columns?         true
           :align-single-column-lines? true})))
+  (testing "map with first-position wrapped value - :align-single-column-lines? true"
```
weavejester: Please ensure all lines are 80 characters or under, as per the contributing guidelines.

sirmspencer:
@weavejester Sorry, I forgot the commit and line width rules.  They are updated.  for the comment on `some->`

the some-> form is not symantically the same.  when z/right* or z/left* returns nil (nothing there, or first element), some-> returns nil which is falsy, but nil should be a passing condition.  The form below works as well.  This change copied the existing logic block, but either is fine with me.

```
  (let [no-break? (complement line-break?)]
    (not (or (some-> zloc z/right* (skip-whitespace-and-commas z/right*) no-break?)
             (some-> zloc z/left*  (skip-whitespace-and-commas z/left*)  no-break?)))))
```

weavejester:
Ah, right. In which case, keep it the way it is. I think that's clearest.

PR #425, opened 2026-07-24
add max-column-anchor-gap to align columns to the gap instead of collapsing on row with short keys

sirmspencer:
@weavejester I can't remember if you want an issue per PR, so I included the details here.  I can make it an issue if you prefer.

### LLM notice
I have reviewed all of what is being submitted.  The PR description was partially generated with llm to show the current and expected.

## Problem

Here, I am not sure if this is a bug or a differene in expected behavior so I went with a new flag.

When `:max-column-alignment-gap` is set, the alignment anchor is still the longest key in the form. Individual rows whose gap would exceed the limit fall back to a single space, but rows with medium-length keys are still extended all the way to the inflated anchor column. This produces inconsistent alignment: the shortest key falls back, medium keys over-extend, and the longest key (which caused the problem) sets the column for all of them.

## Current behavior

Config: `{:align-form-columns? true, :max-column-alignment-gap 5}`

Input:
```clojure
(let [a 1
      abcde 2
      abcdefg 3
      long-name 4]
  [a abcde abcdefg long-name])
```

Output — `long-name` stays as anchor; `a` falls back (gap 9 > 5); `abcde` and
`abcdefg` extend unnecessarily to the `long-name` column:
```clojure
(let [a 1
      abcde     2
      abcdefg   3
      long-name 4]
  [a abcde abcdefg long-name])
```

The same behavior affects maps with `:align-map-columns?`.

## Solution

This PR does not change the behavior of `:max-column-alignment-gap`. Instead it introduces a new flag, `:max-column-anchor-gap`, which applies the limit at anchor selection time rather than per row.

When `:max-column-anchor-gap N` is set, cljfmt selects the largest key for which every shorter key would need at most N spaces to align. Keys longer than the chosen anchor are not padded.

## Expected behavior

Config: `{:align-form-columns? true, :max-column-anchor-gap 5}`

```clojure
;; long-name (gap 8 from a) and abcdefg (gap 6 from a) exceed limit;
;; anchor demotes to abcde (gap for a = 4 ≤ 5)
(let [a     1
      abcde 2
      abcdefg 3
      long-name 4]
  [a abcde abcdefg long-name])
```

## Composing both flags

The two flags are orthogonal and can be combined. `:max-column-anchor-gap` runs
first to select a conservative anchor; `:max-column-alignment-gap` then applies
per-row fallback to the result.

```clojure
;; Config: {:align-map-columns? true
;;          :max-column-anchor-gap 5       ; long-key excluded; anchor = :abcde
;;          :max-column-alignment-gap 3}   ; further trims rows with gap > 3
;;
;; :x falls back (gap 4 > 3); :abc aligns (gap 3 = limit); :abcde is anchor
{:x "a"
 :abc   "b"
 :abcde "c"
 :long-key "d"}
```

## Changes

- New option `:max-column-anchor-gap` added to `default-options` (nil by default)
- `column-start-position` applies cascade demotion when `:max-column-anchor-gap` is set
- `pad-to-position` gains a guard for keys wider than the demoted anchor, preventing
  their spacing from collapsing to zero
- README documents the new option and its relationship to `:max-column-alignment-gap`
- New `test-max-column-anchor-gap` deftest with form, map, and compose cases

## Comparison

https://github.com/user-attachments/assets/bd613ef0-1a6f-4b95-8321-541cabb82193

### DIFF
--- a/cljfmt/src/cljfmt/core.cljc
+++ b/cljfmt/src/cljfmt/core.cljc
@@ -383,6 +383,7 @@
    :align-single-column-lines?            false
    :aligned-forms                         default-aligned-forms
    :max-column-alignment-gap              nil
+   :max-column-alignment-width            nil
    :blank-line-forms                      blank-line-forms
    :blank-lines-separate-alignment?       false
    :extra-aligned-forms                   {}
@@ -670,13 +671,17 @@
 (defn- count-spaces [zloc]
   (if (space? zloc) (node-str-length zloc) 0))
 
-(defn- pad-to-position [zloc start-position {max-gap :max-column-alignment-gap}]
-  {:pre [(or (nil? max-gap) (pos-int? max-gap))]}
-  (let [delta (- start-position (margin zloc))]
-    (if max-gap
-      (let [old-gap (count-spaces (z/left* zloc))
-            new-gap (+ old-gap delta)]
-        (pad-node zloc (if (> new-gap max-gap) (- 1 old-gap) delta)))
+(defn- pad-to-position
+  [zloc start-position {max-gap     :max-column-alignment-gap
+                        align-width :max-column-alignment-width}]
+  {:pre [(or (nil? max-gap) (pos-int? max-gap))
+         (or (nil? align-width) (pos-int? align-width))]}
+  (let [old-gap (count-spaces (z/left* zloc))
+        delta   (- start-position (margin zloc))
+        new-gap (+ old-gap delta)]
+    (if (or (< new-gap 1)
+            (and max-gap (> new-gap max-gap)))
+      (pad-node zloc (- 1 old-gap))
       (pad-node zloc delta))))
 
 (defn- edit-column [zloc column f]
@@ -707,17 +712,24 @@
           (recur (z/right* zloc) 0 acc)
           (recur (z/right* zloc) (inc col) (f zloc col acc)))))))
 
-(defn- column-start-position [zloc col opts]
-  (let [reduce-fn (if (:blank-lines-separate-alignment? opts)
+(defn- column-start-position
+  [zloc col {align-width    :max-column-alignment-width
+             align-singles? :align-single-column-lines?
+             separate?      :blank-lines-separate-alignment?}]
+  (let [reduce-fn (if separate?
                     reduce-column-group
                     reduce-columns)
-        maximizer (fn [zloc c max-pos]
+        collector (fn [zloc c positions]
                     (if (and (= c (dec col))
-                             (or (:align-single-column-lines? opts)
+                             (or align-singles?
                                  (not (single-column-line? zloc))))
-                      (max max-pos (node-end-position zloc))
-                      max-pos))]
-    (inc (reduce-fn zloc maximizer 0))))
+                      (conj positions (node-end-position zloc))
+                      positions))
+        positions (reduce-fn zloc collector [])
+        threshold (when align-width (+ (apply min 0 positions) align-width))]
+    (inc (apply max 0 (if threshold
+                        (filter #(<= % threshold) positions)
+                        positions)))))
 
 (defn- align-one-column
   [zloc col {:keys [blank-lines-separate-alignment?] :as opts}]
--- a/README.md
+++ b/README.md
@@ -316,6 +316,29 @@ In order to load the standard configuration file from Leiningen, add the
    :as everything}                          ;; 27 spaces, falls back to 1 space
   ```
 
+* `:max-column-alignment-width` -
+  a positive integer that limits which keys participate in column
+  alignment. Keys whose end column is at or within N are aligned
+  together; keys that end past column N are not padded and receive a
+  single space before their value. Defaults to nil (no limit).
+  **Experimental.**
+
+  ```clojure
+  ;; Input:
+  {:a 1
+   :abcde 2
+   :abcdefg 3
+   :long-name 4}
+
+  ;; With :align-map-columns? true and :max-column-alignment-width 7:
+  ;; :a and :abcde end at or within column 7 and align together;
+  ;; :abcdefg and :long-name end past column 7 and are not padded
+  {:a     1      ;; ends within column 7, padded to value column
+   :abcde 2      ;; ends at column 7, sets the value column
+   :abcdefg 3    ;; ends past column 7, 1 space
+   :long-name 4} ;; ends past column 7, 1 space
+  ```
+
 * `:blank-lines-separate-alignment?` -
   true if cljfmt should treat blank lines as separators when aligning
   columns. When enabled, alignment groups are separated by blank lines,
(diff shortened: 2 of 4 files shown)


weavejester:
If I'm understanding this correctly, you're proposing a maximum column width size? Anything above the maximum width is separated by just a single space; anything equal to or below is padded as normal. Is that correct?

sirmspencer:
> If I'm understanding this correctly, you're proposing a maximum column width size? Anything above the maximum width is separated by just a single space; anything equal to or below is padded as normal. Is that correct?

Yes, that is correct.    Its specifically this current behavior that I find backwards:

```
(let [a 1
      abcde     2
      abcdefg   3
      long-name 4]
  [a abcde abcdefg long-name])
```

weavejester:
In which case I think we want to call this something like `:max-column-alignment-width`, because this governs the maximum width of the column that should be aligned to. The "anchor" terminology you introduced to describe this feature was a little confusing to me, as you didn't define what exactly you meant by "anchor".

sirmspencer:
> In which case I think we want to call this something like `:max-column-alignment-width`, because this governs the maximum width of the column that should be aligned to. The "anchor" terminology you introduced to describe this feature was a little confusing to me, as you didn't define what exactly you meant by "anchor".

That sounds good to me.   I was trying to stay with the "gap" term since the flags are related.  The "anchor" would be the longest key that is dragging the gap bigger.  Max anchor gap would be the max the anchor could add a gap.  `:max-column-alignment-width` is better, it keeps the "max-column-alignment" part, I'll update.

sirmspencer:
@weavejester I added a text animation video to the description for clarity

cljfmt/src/cljfmt/core.cljc
```diff
@@ -709,16 +712,26 @@
...
+        positions  (reduce-fn zloc collector [])
+        anchor     (when (seq positions)
+                     (if align-width
+                       (let [min-pos   (apply min positions)
+                             threshold (+ min-pos align-width)]
+                         (->> positions
+                              (filter #(<= % threshold))
+                              (apply max)))
+                       (apply max positions)))]
+    (inc (or anchor 0))))
```
weavejester: I think we can make this a little more concise/neater:

```clojure
(defn- column-start-position
  [zloc col {align-width    :max-column-alignment-width
             align-singles? :align-single-column-lines?}] 
  (let [reduce-fn  (if (:blank-lines-separate-alignment? opts)
                     reduce-column-group
                     reduce-columns)
        collector  (fn [zloc c positions]
                     (if (and (= c (dec col))
                              (or align-singles? 
                                  (not (single-column-line? zloc))))
                       (conj positions (node-end-position zloc))
                       positions))
        positions  (reduce-fn zloc collector [])]
    (inc (apply max 0 (if align-width
                        (let [threshold (+ (apply min 0 positions) align-width)]
                          (filter #(<= % threshold) positions))
                        positions)))))
```

(Assuming I haven't made any mistakes)

sirmspencer: @weavejester This works, thanks.  This code changes the int value for max-column-alignment-width to be absolute.  This is good.  

My original function was from the shortest key, which matched the original max-column-anchor-gap.

weavejester: > This code changes the int value for max-column-alignment-width to be absolute.

Sorry, it does? How so?

sirmspencer: My original function was adding off the value of the shortest key (gap).  Since I took your naming and code suggestion, its now calculating the max padding from the start of the keys instead of from the end of the shortest key.  Your way is totally better and more intuitive to what `max-column-alignment-width` should represent.  Forgive me if "absolute" was the wrong word.

sirmspencer: Sorry if that didn't make sense, the final result of this flag is described in the readme change.  The max a column will be padded is calculated from the start of the beginning of the key, to the beginning of the value.

weavejester: Ahh, I see! I was thinking absolute in the mathematical sense, and I was confused at how there could be negative alignment :)

cljfmt/src/cljfmt/core.cljc
```diff
@@ -708,17 +711,23 @@
...
+                             (or align-singles?
                                  (not (single-column-line? zloc))))
-                      (max max-pos (node-end-position zloc))
-                      max-pos))]
-    (inc (reduce-fn zloc maximizer 0))))
+                      (conj positions (node-end-position zloc))
+                      positions))
+        positions (reduce-fn zloc collector [])]
+    (inc (apply max 0 (if align-width
+                        (filter #(<= % (+ (apply min 0 positions) align-width)) positions)
```
weavejester: This recalculates the minimum each time, so N^2 rather than 2N. I think the `let` to define the threshold for the whole filter is necessary (and it ensures the line doesn't overrun in length).

weavejester:
Other than the previous comment, this looks fine. Once that comment is addressed I think we're ready to merge.

sirmspencer:
@weavejester Updated.  I have one more related rule to follow up, if you want to wait to release them all in a group.  I'm going to add a rule to break the link if it exceeds this max width rule.

weavejester:
No problem, I can wait.

This PR looks good, but I've noticed that the `:max-column-alignment-width` option hasn't been added to `cljfmt.main` as a command-line option (`--max-column-alignment-width`). Could you quickly add that in, then I'll merge.

sirmspencer:
> No problem, I can wait.
> 
> This PR looks good, but I've noticed that the `:max-column-alignment-width` option hasn't been added to `cljfmt.main` as a command-line option (`--max-column-alignment-width`). Could you quickly add that in, then I'll merge.

@weavejester I added that in.  Thanks for reviewing!

cljfmt/src/cljfmt/core.cljc
```diff
@@ -671,13 +671,16 @@
...
 
-(defn- pad-to-position [zloc start-position {max-gap :max-column-alignment-gap}]
-  {:pre [(or (nil? max-gap) (pos-int? max-gap))]}
-  (let [delta (- start-position (margin zloc))]
-    (if max-gap
-      (let [old-gap (count-spaces (z/left* zloc))
-            new-gap (+ old-gap delta)]
-        (pad-node zloc (if (> new-gap max-gap) (- 1 old-gap) delta)))
+(defn- pad-to-position [zloc start-position {max-gap     :max-column-alignment-gap
+                                             align-width :max-column-alignment-width}]
```
weavejester: Could you wrap this at 80 characters? i.e.:

```clojure
(defn- pad-to-position
  [zloc start-position {max-gap     :max-column-alignment-gap
                        align-width :max-column-alignment-width}]
```

sirmspencer: yes.  sorry again, i don't normally do this so I forget.

sirmspencer: @weavejester Updated

sirmspencer: wouldn't it be nice if there was a cljfmt setting for this in your own repo :P

weavejester: I had the exact same thought when I wrote the comment. At some point I need to get around to solving that.

PR #429, opened 2026-09-12
Fix missing newline diff output

endvvell:
Previously, `cljfmt check` command failed to print a diff when the only formatting change was appending a trailing newline (such as when `:normalize-newlines-at-file-end?` is set to `true`).

cljfmt_before

This occurred because `clojure.string/split` discarded trailing empty strings, causing DiffUtils to see identical string vectors.

Here is the code demonstrating this effect of the `clojure.string/split` (evaled in the `cljfmt.diff` ns):

```clojure
(comment
  (do
    ;;
    ;; `(str/split s #"\n")` inside `lines` always cuts away the last "\n",
    ;; so the `revised` inside `unified-diff` doesn't have the "\n" at the end
    ;; when needing to produce diffs, and as a result, the diffs do not show
    ;; that a newlines is missing and that it will be added
    ;;
    (assert
     (= (str/split "some string\n" #"\n")
        ["some string"]))
    (assert
     (= (str/split "some string" #"\n")
        ["some string"]))
    ;;
    ;; We could choose to adjust `str/split` to have `-1` limit,
    ;; which, for a string like "some string\n" basically says:
    ;;
    ;; "I found the separator \n. Before that separator I found "some string".
    ;;  And after it is absolutely nothing."
    ;;
    (assert
     (= (str/split "some string\n" #"\n" -1)
        ["some string" ""]))
    ;;
    ;; Because `-1` forces it to keep trailing empty strings,
    ;; it returns that "nothing" as "", resulting in ["some string", ""],
    ;; effectively "preserving" the \n at the end, which is now represented as "",
    ;; which will again "reappear" when `(str/join "\n" ss)` inside `unlines` is executed.
    ;;
    (assert
     (= (str/join "\n" ["some string" ""])
        "some string\n"))
    ;;
    ;; But going for this solution would require adding a third `cond` branch
    ;; in the `annotate-missing-nl`, and altering the second branch to account
    ;; for the effect of another function instead of applying changes in-place.
    ;; Which would look like this:
    ;;
    (defn- annotate-missing-nl
      [file-a-content
       file-b-content]
      (let [file-a-lines (lines file-a-content)]
        (cond
          ;; if file-a is missing a newline:
          (missing-nl? file-a-content)
          ;; appending the null-character marker to the last existing line of file-a,
          ;; to make `DiffUtils` see it as a modified line:
          ;; (this is done to closer imitate git-like diff
          ;;  of the last existing line shown in green when affected)
          (update file-a-lines (dec (count file-a-lines)) str missing-nl-marker)

          ;; if file-a ends with a newline, but file-b doesn't:
          (and (str/ends-with? file-a-content "\n") (missing-nl? file-b-content))
          ;; no need to alter either file:
          ;; (this keeps the trailing "" produced by
          ;;  the `str/split` with `-1` limit arg
          ;;  when the "\n" character was present in `file-a-lines`,
          ;;  which will make `DiffUtils` show a `+` empty line diff)
          file-a-lines

          ;; if both files correctly end with a newline:
          (str/ends-with? file-a-content "\n")
          ;; removing the trailing "" produced by
          ;; the `str/split` with `-1` limit arg,
          ;; so it doesn't affect the produced diff:
          (pop file-a-lines)

          :else file-a-lines)))))
```

This commit introduces a step that detects missing newlines and appends a hidden marker to the last line of the file which does not contain a newline, so the that line is shown as modified in the diff. The new logic then also adds an empty line to the file which did not contain a newline, so that the green diff shows that a newline is added.

The hidden marker is translated into the standard GNU diff "\ No newline at end of file" line. To closer imitate the diff output of the `git diff` command, and the platforms like gitlab (perhaps github as well), with only exception of actually showing the added newline in the diff in green.

Git:

git_example

Gitlab:

example_of_gitlab_newline_diff

Now the diff output of the `cljfmt check` command for appended newlines looks like this:

cljfmt_after_git-like

- I've read through CONTRIBUTING.md and believe to have followed all the mentioned rules.
- I've tested `cljfmt check` command with other formatting rules specified in the cljfmt config file while various code files contained various detected formatting issues, and the diff output related to those issues appears to be unaffected by this change.
- Running `lein test-all` within the `/cljfmt/cljfmt` directory produces "0 failures, 0 errors." result.


weavejester:
Thanks for writing up the issue and the PR. My understanding is that if we have a file `foo` and a file `foo\n`, then the resulting diff should look like:

```diff
--- a/a.txt
+++ b/b.txt
@@ -1 +1 @@
-foo
\ No newline at end of file
+foo
```

As this is the output of `git diff` and `diff -u`. However, even if we fix `lines` like so:

```clojure
(defn- lines [s] (str/split s #"\n" -1))
```

It produces the output:

```diff
--- a/foo.txt
+++ b/foo.txt
@@ -1,1 +1,2 @@
 foo
+
```

Which may be less clear at a glance than the output produced by git or diff.

However, instead of trying to work around this issue in cljfmt, my suggestion is that we instead swap from DiffUtils to [borkdude's text-diff library](https://github.com/borkdude/text-diff) (which is already used for the Babashka variant of cljfmt), and then add this as an option to that library.

What do you think, @borkdude ? Would you be amenable to a PR that adds an option to `borkdude.text-diff/unified-diff` that would make it mimic the output of the CLI diff tools for reporting on ending newline diffs?

borkdude:
Sure!

